Issue #8, opened 2024-03-23
Static Page og:image head tag


Mcrich23:
How do I generate the og:image head tag for a static page?

Mcrich23:
You just add it to the body, nvm

Mcrich23:
It compiles, but doesn't enter the head. Would love to see a way to add onto the default heading.

twostraws:
Great fix – thank you!

PR #9, opened 2024-03-23
Add image to socials for Static Page

Mcrich23:
I have added a new variable to the protocol `StaticPage` called `image` that is an optional url to the desired image for social sharing. It hooks into rendering to plug the image url into the `Page` initialized. By default, it's set to nil.

A fix for issue #8


twostraws:
Thank you! I might come back to this in the future – if we detect a favicon in the Assets folder, perhaps we should just use it automatically.

Issue #1, opened 2024-03-22
Add a CLI tool to create new sites and launch them in a local server


Mcrich23:
I would love to be able to run the target and it opens `index.html` properly with formatting. Right now, I build it and open it locally and get this:

<img width="1624" alt="Screenshot 2024-03-22 at 1 27 51 PM" src="https://github.com/twostraws/Ignite/assets/81453549/c9f6f48e-810b-48b4-a4d3-40f08709008e">

But running it on a cloudflare site works fine

twostraws:
Most folks are likely to do this using a local web server. The most common was is using Python, which ships with one. `cd` into your Build directory on the command line, then run `python3 -m http.server` to start the server. You should now be able to visit http://localhost:8000 to access your site, rendered normally.

Mcrich23:
Thank you, I figured this out. Might be something to add to the executable to have it built in.

twostraws:
Yes, I agree. This issue should stay open, but I've renamed it to the bigger problem: we need a way to generate new sites quickly, and also run those sites.

Mcrich23:
@twostraws I could write shell scripts to do these things. However, I don't know how to turn this into a cli.

Mcrich23:
@twostraws I have put together a couple shell script to do exactly what it needed but you should probably be the one to publish it. It has 2 commands: 

`ignite create-template [-n|--name new_folder_name] [destination_path]` which fetches the starter repo and unzips it and if you specify a name it will rename IgniteStarter to the name.

and `ignite run [-p|--port] [-d|--directory]` which will startup a server on port 80 and run the package
[Ignite-CLI.zip](https://github.com/twostraws/Ignite/files/14733859/Ignite-CLI.zip)

Mcrich23:
There were some bugs, so I made fixes to the script: 
[Ignite-CLI.zip](https://github.com/twostraws/Ignite/files/14810290/Ignite-CLI.zip)

twostraws:
This is now resolved.

PR #17, opened 2024-03-29
Added local server functionality

markbattistella:
Adds in functionality to serve a local version of the Ignite website when run directly from Xcode. (Addresses issue #1)

### What this change does

1. Adds a new `testingPort` parameter to the `Site` protocol and extension.
2. Adds a new function (`serveLocalSite(on:)` to the PublishingContext
3. Adds that new function to the `publish()` method
4. Creates a new `Server` struct to serve the website directly from Xcode

### How it works

The `Server` struct has one public method - `start()` which firstly checks if the port intended for use - default `8000` - is available. If it isn't then an error is thrown.

If that port is available, then an AppleScript is called to launch a Terminal window with the `cd /Build && python3 -m http.server 8000` call.

It then launches the webpage after a 2 second delay.

On subsequent Xcode runs, the AppleScript does not fire or open new Terminal windows.

### Customisation

The use of the `testingPort` in the `Site` struct allows the user to specify a custom port for use in the event they or some other local service is using port `8000`.

### Considerations

This obviously relies on two external areas - Terminal and AppleScript. Neither should be viewed as high level risks, however there could be additional security alerts or notifications in future macOS releases, or even a deprecation of AppleScript.

Though I doubt users would be removing Terminal from their system entirely (or if even possible) this does mean that it will launch for the server opposed to other preferred apps. Not an issue, but something which may come up as a point of contention from die-hard terminal fans.

### Future expansion

Without feature creep (into the `Site` struct) or over-engineering a flag to allow the user bypass or disallow webpage serving.

Something along the lines of `--env=development` vs `--env=production`.

This way if we were to implement a Github Action or another external script to generate websites, then we don't fall into the trap of it attempting to launch Terminal or a web browser.

### DIFF
--- a/Sources/Ignite/Framework/Server.swift
+++ b/Sources/Ignite/Framework/Server.swift
@@ -1,0 +1,120 @@
+//
+// Project: Ignite
+// Author: Mark Battistella
+// Website: https://markbattistella.com
+//
+
+import Foundation
+
+/// Errors that can be thrown by the `Server` struct.
+public enum ServerError: LocalizedError {
+
+    /// Indicates that the server is already running on the specified port.
+    case serverAlreadyRunning
+
+    /// Indicates a failure to execute an AppleScript, including the error message.
+    case appleScriptExecutionFailed(String)
+
+    public var errorDescription: String? {
+        switch self {
+            case .serverAlreadyRunning:
+                "Server is already running"
+            case .appleScriptExecutionFailed(let message):
+                "AppleScript failed to execute. Error message: \(message)"
+        }
+    }
+}
+
+/// Represents a server that can serve content from a directory over HTTP.
+public struct Server {
+
+    /// The port on which the server should run.
+    public let port: Int
+
+    /// The file URL of the directory whose contents should be served.
+    public let directoryPath: URL
+
+    /// The path of the directory as a string, with spaces escaped for shell commands.
+    private let directory: String
+
+    /// Initializes a new server with the given port and directory path.
+    ///
+    /// - Parameters:
+    ///   - port: The port on which to serve the directory's contents.
+    ///   - directoryPath: The file URL of the directory to serve.
+    public init(port: Int, directoryPath: URL) {
+        self.port = port
+        self.directoryPath = directoryPath
+        self.directory = directoryPath.path(percentEncoded: false).escapingSpaces
+    }
+
+    /// Starts the server, throwing an error if it cannot start.
+    ///
+    /// This method attempts to start serving the contents of `directoryPath` over HTTP on `port`.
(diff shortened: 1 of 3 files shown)


Mcrich23:
I think this is great! and if @twostraws does not want to publish the cli I sent in #1, I am more than happy to publish it with homebrew.

Mcrich23:
One thing, I have a github action that builds and published my site, can you please clarify how I would tell it to only build the site and not serve it?

markbattistella:
> One thing, I have a github action that builds and published my site, can you please clarify how I would tell it to only build the site and not serve it?

Not yet a function in this PR. Trying to think of the easiest way without having user interaction.

I think the optimal way would be so that it doesn't require schemas or anything too complicated - particularly for the simple nature of this package.

The only way I can think of it now would be adding in another flag into `Site` of an enum for the website state:

```swift
enum SiteState {
  case development // build and run local server
  case production  // build only
}
```

Then in the `start()` function we can check the state and only open web browsing and terminal if in `.development`.

But then it opens the world of deploying, forgetting to change it, action issues, etc. Or even the vice-versa.

Ideally, your CLI version would be the robust method but takes away the ease from users who are learning Xcode Swift development, but then need to learn terminal.

I think this really does dial down to how @twostraws and team want to aim this type of package. Is it for beginners with everything inside the one package, or advanced where terminal and running the python server is a prerequisite.

Mcrich23:
maybe call it production or maybe cli mode?

twostraws:
There are two very different approaches here, and I like parts of both. @Mcrich23's approach is a command-line tool, which neatly separates creating new projects / running them from the process of building sites, and @markbattistella's approach puts about half the work in Swift, combining running with building.

I'm going to read through all your code now, and try to produce some kind of composite that brings together the best of both of them.

twostraws:
Okay; I've pushed up a command-line tool. You should be able to install it by cloning this repository then running `make`, or perhaps `sudo make` depending on your directory permissions. If you then run `ignite` it should show you how to use the tool.

Hopefully this takes the best of both of your approaches – let me know if you're happy to close this PR. If you have further improvements etc, that should probably be fresh issues/PRs.

Mcrich23:
Can we please also get it on Homebrew?

twostraws:
Homebrew is definitely possible, but one thing at a time please – does this command-line tool achieve what you both were looking for?

Mcrich23:
Yes! the one thing I would recommend is being able to change the template name on when running the command

twostraws:
I've added a --template option now; hopefully that's what you're looking for! If so, I can close this issue, and formally document the CLI tool.

twostraws:
I haven't heard back about this, so I'm going to close the issue now and document the CLI tool. Thank you for your help!

PR #19, opened 2024-04-01
Added favicon support for safari

Mcrich23:
As with favicon support before (#8, #9), safari was not supported. This pr should fix that.

### DIFF
--- a/Sources/Ignite/Elements/Head.swift
+++ b/Sources/Ignite/Elements/Head.swift
@@ -61,6 +61,7 @@ public struct Head: HTMLRootElement {
             
             if let favicon = context.site.favicon {
                 MetaLink(href: favicon, rel: .icon)
+                MetaLink(href: favicon, rel: "apple-touch-icon")
             }
         }
 


twostraws:
This has been sitting waiting review for a while, which is my fault – I got into a bit of a rabbit hole on what exactly modern favicon support looks like, and in doing so even updated <https://www.hackingwithswift.com> to include the latest recommended approach. It's hard to get a great answer because so much just seems to be 10 years old, but from what I can see a good approach is:

1. Using an SVG for the main favicon
2. Using a 180x180 PNG for the Apple touch icon

I have no idea whether using 1 automatically makes 2 work, because if it doesn't it really ought to!

However, I'm wondering whether a better solution might work here: can we detect what favicon files they have in their Assets folder, and use that to generate all the correct tags automatically?

Mcrich23:
we can. I was looking at this as well and apple requires an extra metatag for safari. Thats what this pr is for. If you want me to, I can set it to auto look for favicon.png/svg/jpeg, and then people can override it if they choose.

aduryagin:
> apple requires an extra metatag for safari

Worth to mention that **apple-touch-icon** has been deprecated. Apple requires this only for Safari versions prior to 15.4
https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes#Web-API
https://developer.mozilla.org/en-US/docs/Web/Manifest/icons#browser_compatibility

Mcrich23:
@twostraws Do you want to merge this, or is it obsolete? I have been away from Ignite for a bit...

cpah:
I am still using the original release (0.0.2) of Ignite and this line in the `MySite struct` in `Site.swift` works just fine for me in Safari and Edge:
```swift
var favicon = URL(string:"/images/favicon.png")
```
I guess that no further action is required?

JPToroDev:
It seems like no further action is needed. I'll close this for the time being. Thanks!

PR #21, opened 2024-04-02
Fix location of CSS elements for default template

jpurnell:
In the default template, the locations of CSS files are prefixed with a "/" This modifies the MetaLink file to remove this prefix

### DIFF
--- a/Sources/Ignite/Elements/MetaLink.swift
+++ b/Sources/Ignite/Elements/MetaLink.swift
@@ -11,13 +11,13 @@ import Foundation
 /// a stylesheet.
 public struct MetaLink: HeadElement {
     /// The standard CSS you should include on all Ignite pages.
-    public static let standardCSS = MetaLink(href: "/css/bootstrap.min.css", rel: "stylesheet")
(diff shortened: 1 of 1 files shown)


twostraws:
Hello! Thank you for this. However, I believe the leading slash is correct. Why would you want to remove it?

jpurnell:
When I run a simple project with the default template, if you open the index.html file from the build folder, the browser can’t find the css files’ relative location (at least on my local machine), and I have to remove the leading slash. Maybe we change the build settings to make that leading slash toggle-able?

twostraws:
Broadly speaking, most sites aren't server up as local files, and are instead provided from web servers. If you want to run a local server for testing purposes, open your Terminal app then `cd` into the Build directory and run `python3 -m http.server`. We are currently trialing a command-line tool to make this a little easier (https://github.com/twostraws/Ignite/pull/17), but that's not fully delivered yet.

twostraws:
This repository now includes a CLI tool that makes it easier to run a local web server for testing, so I'm going to close this PR.

PR #26, opened 2024-04-16
Move the trailing <script> chunks to the end of <body> from  <html>

steve-h:
I moved the code and tested in my own copy of samples repo. Things seem to work. The testability is not obvious to me in the Ignite repo, but the existing tests pass.

I considered loading the pure scripts in the head section with a defer parameter but the tool tip script snippet looked like it needed to run last so I did not try that.


twostraws:
Perfect – thank you!

PR #30, opened 2024-04-25
Added 'text-decoration': .none, .through and .underline for text

Bryant-24:
I would like to add the ability to change the text formatting using the `textDecoration` modifier, which has the values: 
* `.underline` 
* `.through`
* `.none`
Screenshot 2024-04-24 at 5 45 04 PM
<br><br>
The `.none` value can also be used for links if you want to place a link without the underscore


twostraws:
Wonderful – thank you!

Issue #31, opened 2024-04-27
Feature Request: allow for custom positioning of Card Content Overlay


TheApApp:
Positioning of overlay

```
Card(imageName: "/images/wind.jpg") {
                Text("Arya, Samoyed")
                    .foregroundStyle(.white)
                    .fontWeight(.bold)
            }
            .contentPosition(.overlayCenter) // possible solution? 
```

pd95:
I was looking into this problem and found the following solution to position the text in the middle of the card:
1. Make the text take up maximum space
2. Add the two classes "text-center" (for horizontal-alignment) and "align-content-center" (for vertical-alignment)

```swift
Card(imageName: "/images/wind.jpg") {
    Text("Arya, Samoyed")
        .foregroundStyle(.white)
        .fontWeight(.bold)
        .frame(width: "100%", height: "100%")
        .class("text-center", "align-content-center")
}
.contentPosition(.overlay)
```

Now it would probably be nice to fit everything into a nice API... something like a `ZStack` with alignment properties 😃

pd95:
You are right: `.overlay(alignment:)` would solve the various alignments, without the need for a `ZStack` 😉
I could prepare this, if you like.

twostraws:
Sounds good – although I'd be interested to see how much we can retain code backward-compatibility.

pd95:
I've created a prototype with an additional modifier and a new `enum ContentAlignment`.
```swift
enum ContentAlignment: CaseIterable {
    case topLeading
    case top
    case topTrailing
    case leading
    case center
    case trailing
    case bottomLeading
    case bottom
    case bottomTrailing
}

extension Card {
    func contentAlignment(_ newAlignment: ContentAlignment) -> Self {
        // more code ...
    }
}
```

The positioning in the overlay is triggered by the Bootstrap CSS classes for [text alignment](https://getbootstrap.com/docs/5.3/utilities/text/#text-alignment) and [vertical alignment](https://getbootstrap.com/docs/5.3/utilities/vertical-align/). (=the same properties used for the current implementation of `.overlayCenter`)

<img width="1033" alt="Bildschirmfoto 2024-04-29 um 21 52 21" src="https://github.com/twostraws/Ignite/assets/25307503/57b14c73-962f-455e-b23a-fe56b285c5d4">

```swift
Section {
    for alignment in Card.ContentAlignment.allCases {
        let alignmentName = String(describing: alignment)
[...]

PR #32, opened 2024-04-27
Enhancing `Card`

pd95:
Inspired by your HWS+ session about Ignite today, I was looking more into how `Card` is working and how the generated HTML relates to the [Bootstrap documentation](https://getbootstrap.com/docs/5.3/components/card/#image-caps).
I've seen that it would be possible to place the image above and below the content text and that Bootstrap is "capping" the edges of the image accordingly. The current version of `ContentPosition` does only support placing the content below the image (=first image, then content) and hardcoded the CSS classes accordingly to "card-img-top".

I've enhanced this by adding additional options. 
Further I added a fix to allow the image being placed into content of a `Card`: The image has to get the "card-img" CSS class for proper sizing.

I'm not yet fully convinced by the current solution. Do you think `ContentPosition` enum needs more adjustments?
Is `ContentPosition` the correct wording? Wouldn't `ImagePosition` `.top`, `.bottom` and `.background` be more appropriate?

Screenshot showing all 4 variations

### DIFF
--- a/Sources/Ignite/Elements/Card.swift
+++ b/Sources/Ignite/Elements/Card.swift
@@ -22,24 +22,31 @@ public struct Card: BlockElement {
     }
 
     /// Where to position the content of the card relative to it image.
-    public enum ContentPosition: String, CaseIterable {
+    public enum ContentPosition: CaseIterable {
         /// Positions content below the image.
-        case bottom = "card-img-top"
+        case bottom
 
         /// Positions content above the image.
-        case top = "card-img-bottom"
+        case top
 
         /// Positions content over the image.
-        case overlay = "card-img-overlay"
+        case overlay
 
         public static let `default` = Self.bottom
 
         var imageClass: String {
-            self == .overlay ? "card-img" : rawValue
+            switch self {
+            case .bottom:
+                "card-img-bottom"
+            case .top:
+                "card-img-top"
+            case .overlay:
+                "card-img"
+            }
         }
 
         var bodyClass: String {
-            self == .overlay ? rawValue : "card-body"
+            self == .overlay ? "card-img-overlay" : "card-body"
         }
     }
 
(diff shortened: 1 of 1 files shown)


twostraws:
First: thank you for this contribution! Could you share the Ignite code used to generate your examples, please? It helps everyone see how this all looks at the call site.

Second: I'm certainly not committed to the current spelling, and am happy to bounce around others until we find something that works well 🙂  Generally speaking, what I'm keen to avoid are modifiers that mean different things in different places, and in this case my first preference would have been `cardStyle()` because it's clear that this is a `Card`-specific modifier. Sadly, that was taken by existing card styling code, so I ended up with `contentPosition()`. `contentPosition()` seemed a little less common to me than `imagePosition()`, but both aren't used so both seem sensible choices. (Note: If this changes we should retain the existing `contentPosition()` modifier and mark it as deprecated, to avoid breaking code.)

Wherever we end up, it would be helpful to make sure it aligns well with any future work done to accommodate #31.

pd95:
The change I've added is just a possible implementation of `.overlayCenter`.
I'm wondering whether it would make more sense to work on `HStack`, `VStack` and `ZStack` for more control over positioning, instead of building some alignment options into each possible supported element. 

@twostraws : You have probably have some idea on how to handle the alignment options or even did some experimentation with stacks already?

twostraws:
Thank you for clarifying how the code is used! I'll merge this change in now. There's still scope to discuss spelling of this API in the future, but at least folks can benefit from this now rather than waiting further.

SwiftUI's stacks are tricky, because they make a promise I don't particularly want to keep 😅

If you say `HStack { View() View() View() }`, SwiftUI will place those views horizontally. That might be fine when you're building for iPhone, but on the web really we'd want that `HStack` to wrap to two lines or even three when space is restricted. That's exactly how `Section` works, which is why I gave it a different name from `HStack` – I don't want folks to consider it a "false friend", in language terms.

*That being said*, we could still add a `VStack` and even a `ZStack`, leveraging parts or all of Bootstrap's [flex utilities](https://getbootstrap.com/docs/5.3/utilities/flex/). That *could* include a `HStack`, effectively making a `Section` that never wraps, but again I think we ought to tread carefully there.

twostraws:
@pd95 When you have a moment, could you open a PR on IgniteSamples showing off the new functionality with some sample code? Thank you!

Issue #33, opened 2024-04-28
Making a site in a subdirectory 


gavineadie:
I'd like to use Ignite to build a sub-site that doesn't live directly under the root URL but under a subdirectory of that.  I've hunted around to see if that is catered for in the existing code base but not found it (which might easily be a result of my not looking hard enough).

I saw the PR "Making the buildDirectory changeable #16" and thought that might be a way to accomplish my goal by setting `PublishingContext(.., buildDirectoryPath="subdir/Build")` but, no.  Similarly, I tried setting `Site.url = https://example.com/subdir` but, no there too.

Building a site and placing all its files into a subdirectory doesn't work because assets, css, etc, being referenced as, `/xxx` are not found because they're actually at `/subdir/xxx`.  I'm going to dig into this more and submit a PR, if necessary, but wanted to check that I hadn't missed something obvious before setting out.

twostraws:
Yes, right now we go back to the root every time, including when referencing things like paths in the `Image` element. 

Perhaps if this were made a property of `Site`, `basePath` or similar, with a default value of `/`, that could be something that's easy to adjust anywhere. To make this less likely to break backwards compatibility, we could perhaps make the default an empty string, then elsewhere simply prepended that to the paths? So, existing code like `Image("/images/whatever")` would remain pointing to the root, whereas someone with an override for `basePath` of `/subsite` would automatically end up with `/subsite/images/whatever`.

Just some thoughts, but it seems eminently soluble.

gavineadie:
Good! .. I'm relieved your suggested approach matches what I was considering!  I read through a lot of your code last night and it seemed that making a change to a property that reaches `render(..)` via `context` would cause the least disruption.  Not every element passes through `render(..)` at the necessary granularity .. but I'll stop surmising here and tackle some code.

twostraws:
Thank you! Let me know if there's anything you need help with.

gavineadie:
Thanks Paul .. A little looking over my shoulder, if you would?

_For others reading along, I'm calling `https://site.com` [its URL has no path] a "root" site, and a site, off the root, at `https://site.com/non-root` is a "non root" site._

**Ignite** generates "root" sites and I would like to allow it to also generate "non root" sites.  I see two places where the author of a new site could make the choice to construct one or the other.  The partial code below shows both:

* set the `buildDirectoryPath` to specify a hierarchy.

```
	@main
	struct IgniteWebsite {
		static func main() async {
			let site = ExploreSite()
[...]

PR #35, opened 2024-04-29
Added path to FeedConfiguration to make RSS endpoint configurable

mfreiwald:
When upgrading a website to use Ignite, it helps to keep the same RSS feed URL as before.
The current implementation has hardcoded the endpoint to `{site_url}/feed.rss`.

This PR adds a `var path: String` to the `FeedConfiguration` to make this endpoint configurable.

It also works with paths like `path: "blog/rss.xml"`

### DIFF
--- a/Sources/Ignite/Components/FeedLink.swift
+++ b/Sources/Ignite/Components/FeedLink.swift
@@ -20,7 +20,7 @@ struct FeedLink: Component {
                         .margin(.trailing, 10)
                 }
 
-                Link("RSS Feed", target: "/feed.rss")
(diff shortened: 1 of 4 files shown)


twostraws:
Thank you! Do you think this might conflict with any work required to make #33 work? If not, I'm happy to merge it in.

mfreiwald:
No, I don't think that it makes more conflicts than with the current implementation.

twostraws:
Okay. I'll merge this now, and we can revisit once #33 moves forward. Thank you!

PR #36, opened 2024-04-29
Fix the reversed `imageClass` for `ContentPosition`

pd95:
While writing the documentation on https://github.com/twostraws/IgniteSamples/pull/8 I found this bug (probably caused by my refactoring): I've flipped the image classes in `ContentPosition.imageClass`.

This bug is showing as incorrectly "capped" corners:
image

This additional change is fixing this.


twostraws:
Thank you!

PR #37, opened 2024-04-29
Allow to specify alignment for Card `overlay`

pd95:
This PR enhances `ContentPosition.overlay` to take an alignment argument: `.overlay(alignment:)`
The alignment is defined as

```swift
enum ContentAlignment: CaseIterable {
    case topLeading
    case top
    case topTrailing
    case leading
    case center
    case trailing
    case bottomLeading
    case bottom
    case bottomTrailing
}
```

The implementation uses the Bootstrap "text alignment" and "vertical alignment" classes.

The following example usage shows all nine cases:

```swift
Text(markdown: """
To control the position of the overlay you can specify an alignment using `.overlay(alignment:)`
with one of the following options:
""")

Section {
    for alignment in Card.ContentAlignment.allCases {
        let alignmentName = String(describing: alignment)
        Card(imageName: "/images/photos/dishwasher.jpg") {
            Text(markdown: "`.\(alignmentName)`")
                .foregroundStyle(.white)
                .backgroundColor(.lightGray)

            Link("Back to the homepage", target: "/")
                .linkStyle(.button)
        }
        .contentPosition(.overlay(alignment: alignment))
        .imageOpacity(0.5)
    }
}
.margin(.top, .large)
.columns(3)
```

All-Alignment-Cases

### DIFF
--- a/Sources/Ignite/Elements/Card.swift
+++ b/Sources/Ignite/Elements/Card.swift
@@ -40,6 +40,7 @@ public struct Card: BlockElement {
         public static let `default` = Self.bottom
         public static let overlay = Self.overlay(alignment: .topLeading)
 
+        // MARK: Helpers for `render`
         var imageClass: String {
             switch self {
             case .bottom:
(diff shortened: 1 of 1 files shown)


pd95:
@twostraws : I've reworked my solution and documented it in #31.
It uses now `.contentPosition(.overlay(alignment: .center))` for centering the overlay. I've added `overlay` as a static variable to `ContentPosition` for backward compatibility. But I removed `.overlayCenter` as this had been included previously "by accident" and was not part of a release.

Let me know whether it needs more rework. I will add a separate PR for IgniteSample later.

twostraws:
This is really excellent work – thank you!

PR #38, opened 2024-04-29
prepare for "non root" sites ..

gavineadie:
In preparation for allowing Ignite to construct "non root" sites (ie: `https://site.com/NONROOT/index.html`), some protection and adjustments:

* issue a warning if there's a `\` in `buildDirectoryPath`
(the warning is emitted in an Xcode build but not a CLI build)
* allow the site directory to be created
(change `withIntermediateDirectories` to `true` in `PublishingContext`)
* don't copy hidden (".") files into site
(I don't think web sites use hidden files, certainly not `.DS_Store` !)

_My first PR -- please suggest any improvements_

### DIFF
--- a/Sources/Ignite/Framework/Site.swift
+++ b/Sources/Ignite/Framework/Site.swift
@@ -159,6 +159,11 @@ extension Site {
     ///   The default is "Build".
     public func publish(from file: StaticString = #file, buildDirectoryPath: String = "Build") async throws {
         let context = try PublishingContext(for: self, from: file, buildDirectoryPath: buildDirectoryPath)
+
+        if buildDirectoryPath.contains("/") {
+            context.addWarning("in \"\(#function)\" buildDirectoryPath is heirarchical (\"\(buildDirectoryPath)\")")
+        }
+
         try await context.publish()
 
         if context.warnings.isEmpty == false {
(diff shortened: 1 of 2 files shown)


twostraws:
Thank you for this! Your first PR is always the most stressful, but it gets easier 🙂

The second change is a good step forward. The third change is interesting, although I'm not sure it's necessarily a good idea – some web servers use hidden files, e.g. `.htaccess` on Apache, to provide custom configurations for websites. If users put these into a directory, it would be annoying if they weren't copied. Perhaps if we copied hidden files, excluding those that appear in a blocklist? (Also, randomly: that method is called `copyResources()`, which is a bit confusing now that I added a separate Resources directory!)

I'm afraid I don't understand what the first change is trying to do 😅

Broadly speaking, it's a good idea to make each pull request solve one distinct problem, unless of course many things are tied together. So, if this were three individual PRs, I'd merge 2 on the spot, and we'd discuss 1 and 3 further. That's okay, though; leave this as one PR for now, and we can figure it out bit by bit.

gavineadie:
I submitted (1) to reduce an accidental use of `a/b` [I just noticed my PR talks about `\` instead of `/`] failing in some nasty way if I submit a buggy or incomplete later PR.  Prior to (2), having a `/` in `buildDirectoryPath` would not generate a site.  It was my intent to remove that warning when I was done.

(3) .. yes, `.htaccess` trumps number (3) so don't use it -- after all, a `.DS_Store` or two isn't a big deal.

Summary: Take (2) and reject (1) and (3)!  As a PR novice should I back out the two, or you just ignore them?

gavineadie:
I removed/reverted (1) and (3) from this branch ..

twostraws:
Nailed it – thank you!

PR #40, opened 2024-05-01
Update Makefile

TheArchitectLabs:
Add command to Makefile to see if the usr/local/bin directory exists before installation. If it does not, it will be created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 install:
 	@echo "Building the Ignite command-line tool...\\n"
 	@swift build -c release --product IgniteCLI
-	@if ! [ -d "/usr/local/bin" ]; then mkdir -p /usr/local/bin; fi
+	@[ $(shell id -u) -eq 0 ] && sudo mkdir -p /usr/local/bin || { echo "You do not have root permissions. Either manually create the /usr/local/bin directory or run as \'sudo\'"; exit 126; }
 	@(install .build/release/IgniteCLI /usr/local/bin/ignite 2> /dev/null && (echo \\n✅ Success! Run \`ignite\` to get started.)) || (echo \\n❌ Installation failed. You might need to run \`sudo make\` instead.\\n)
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 install:
 	@echo "Building the Ignite command-line tool...\\n"
 	@swift build -c release --product IgniteCLI
+	@if ! [ -d "/usr/local/bin" ]; then mkdir -p /usr/local/bin; fi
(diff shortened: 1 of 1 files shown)


twostraws:
This is a big help – thank you! Are we *able* to make that directory, ensuring it has the correct permissions? If not, we might be better off printing an error and exiting – "/usr/local/bin could not be found – please create it, then move .build/release/IgniteCLI to /usr/local/bin/ignite manually", for example.

TheArchitectLabs:
I tested on my mac and MacBook Air and was able to create the bin directory as long as sudo was used. I also tested and created different folders in the same spot and was able to successfully install ignite using the script.

twostraws:
And if sudo *isn't* used? If we're going to create /usr/local/bin, we might want to use `whoami` first, for example, to check whether the user has permission to make it or not.

TheArchitectLabs:
I can run 'whoami' and it will tell me who is logged in but doesn't tell me if I have permissions to the folders as needed. I found an example where you can run 'sudo whoami' but that requires the user to put in their password. Not ideal in my opinion but could work. That would return if they have root or not. If they do, then they could create the /usr/local/bin directory without sudo being needed. Otherwise, it would just default to a message that says for them to create it manually and move .build/release/IgniteCLI to /usr/local/bin/ignite per your original comment.

Thoughts?

I very much appreciate the conversation and education with this. I am really enjoying trying to figure this out. However, I have a side question...

Where can I best learn what commands are available to utilize and how to do this scripting? I wouldn't have known 'whoami' but after you mentioned it, I was able to Google it and see some scripts. I looked through the makefile (didn't know these existed until this project) and I am still trying to figure out the meaning of the @(install line. 

I have found other articles but many refer to python or ubuntu or other languages that I have never learned. I want to learn, just need to know where to start.

Thank you for all your help and comments.

twostraws:
What I was thinking is that if `whoami` returns "root" then we're safe to create the directory; I don't believe sudo would be required for that. However, on reflection I'm wondering whether we should 1) check if the directory exists, 2) if it doesn't, attempt to create it, and 3) if creating it fails, bail out and tell the user they are missing a /usr/local/bin directory. Something like https://stackoverflow.com/a/16633130 ought to help.

As to where you can learn more about the various command-line tools, I have a book called Beyond Code that covers it as 1/4 of the book 🙂   I learned my skills through a much older book called Unix Power Tools.

TheArchitectLabs:
Updated the makefile. If the user can create the /usr/local/bin directory, then it will be created and the installation will complete. If the directory can't be created, the installation will be stopped with an error 126 (permissions).

Thank you so much for the reading information. I am a HWS+ subscriber so I went straight to Beyond Code. I am also going to check out the Unix Power Tools book you mention. With all of this, I am starting to understand a little bit more of what is happening.

twostraws:
This looks perfect to me – thank you!

PR #42, opened 2024-05-03
Add attributes to the script element

mfreiwald:
The Script tag doesn't contain the attributes, but often there is a need for it.
E.g. `<script data-domain="site" src="/js/analytics.js"></script>`

This PR adds the CoreAttributes to the script tag and I added some tests for it.


twostraws:
Great job, and I love the tests too – thank you! 🙌

PR #43, opened 2024-05-03
New element: Abbreviation

henrikac:
This PR add an `Abbreviation` element. `Abbreviation` renders a `<abbr>` tag. This is something I find myself use from time to time and thought it'd be a nice addition to the collection of Ignite elements.

I added a single test for `Abbreviation`.

### DIFF
--- a/Sources/Ignite/Elements/Abbreviation.swift
+++ b/Sources/Ignite/Elements/Abbreviation.swift
@@ -1,0 +1,35 @@
+//
+//  Abbreviation.swift
+//  Ignite
+//  https://www.github.com/twostraws/Ignite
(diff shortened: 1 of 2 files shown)


twostraws:
Thank you! I'm not familiar with the element, but perhaps you know: does the abbreviation being displayed on the page always need to be a simple string, or could it involve bold, italics, or more complex elements?

henrikac:
I usually just use simple strings and therefore I did not think about if the abbreviation text can be formatted e.g. italic or bold. It turns out that it can. I will add the appropriate constructors and tests for this :)

twostraws:
Sounds good to me! I'll merge this now. Thank you!

PR #45, opened 2024-05-07
Changes to allow Ignite to generate a subsite

gavineadie:
Ignite generates a website at the root of a web server: `http://site.com`

This change in the PR allows a site to be created to reside at: `http://site.com/subsite`

The change is driven by simply changing the URL in the `Site` struct

```
struct SampleSite: Site {
    var name = "Sample Ignite Site"
    var url = URL("http://site.com/subsite")    <--
     . . .
```

Unit tests have been added for those HTML elements that are impacted by this change.  Additionally, the **IgniteSamples** site has been generated as such a subsite, then tested for fidelity and a file-by-file comparison with the 'out of the box' version (including `Sitemap.XML` and the RSS feed URL).  That subsite version is at: `https://ramsaycons.com/ignitesamples`

One incomplete aspect of this change is that the `ignite run` command has not been changed to accommodate viewing the subsite [fixing this may be easy enough; I've not tried].  For viewing during development, the content of the Build directory needs to be placed in the correctly named directory of a 'real' web server (which could be one of several simple examples to run locally).

The Ignite CLI version, and the Git tag for this PR, are 0.2.2 ..

### DIFF
--- a/Sources/Ignite/Elements/Image.swift
+++ b/Sources/Ignite/Elements/Image.swift
@@ -94,7 +94,11 @@ public struct Image: BlockElement, InlineElement, LazyLoadable {
     ///   - context: The active publishing context.
     /// - Returns: The HTML for this element.
     private func render(image: String, description: String, into context: PublishingContext) -> String {
-        "<img src=\"\(image)\"\(attributes.description) alt=\"\(description)\"/>"
+        """
+            <img src=\"\(context.site.url.path)\(image)\" \
+            \(attributes.description)\
+            alt=\"\(description)\"/>
+            """
     }
 
     /// Renders this element using publishing context passed in.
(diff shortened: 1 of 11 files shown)


twostraws:
This looks great – thank you! The `ignite run` command is a thorny one. Perhaps we should at least say "this site specifies a custom subfolder, so it can't be previewed locally" until we have a better idea?

Note: Please remove the "Created by…" lines. The best place for those is adding credits in the README, otherwise if someone else comes along and dramatically rewrites your code things get rather messy in terms of credit and responsibility.

gavineadie:
Is finding the String `<link href="/css` in the file `Build/index.html` too horrible a test for a non-subsite?  If acceptable, the following commit implements that.

twostraws:
Any test is better than no test, and we can always return to it and improve later on if needed. So, go for it!

gavineadie:
Unless you've other suggestions, Paul, I'm done with this PR so you could pull it.

twostraws:
Thank you!

Issue #47, opened 2024-05-07
Question: How to add Font Awesome Support in <head>


TheArchitectLabs:
I like to utilize FA in my site. However, the instructions for use are to copy this code into the <head>:  <script src="https://kit.fontawesome.com/xxxxxxxxxx.js" crossorigin="anonymous"></script>

How can I add this to the <head>? I don't see a way to modify that element.

Thank you for the wonderful work on this and I apologize if there is a simple fix that I am just missing.

twostraws:
Right now `Ignite` provides two options for headers:

1. You can build your own from scratch, using `Script` in this case to get the effect you want.
2. You can use the standard headers.

On a live stream about Ignite I mentioned that really we needed some sort of middle ground: "give me all the standard headers, then add these extras…" I'm afraid that isn't currently a thing, but I think this issue is a good reminder that it is needed.

pd95:
I remember the discussion during the live stream. I would volunteer to develop a solution for this, but currently I'm still an "`Ignite` newbie". So I'm wondering what the right approach is to implement it. Here follows my current understanding of the problem:

To include some additional CSS or JS Script file to the website (=every page) one has to to modify the `MyTheme` source which renders the HTML of a page. By default, the content of the <head> tags is created by the call to `Head(for: page, in: context)`. This results in the rendering of the standard headers as they are defined in the `Head` initializer.

To add the additional CSS (using `MetaLink(href: "/css/custom.css", rel: "stylesheet")`) or JS file (using `Script(file: "/js/custom.js")`) we have to use the other initializer (based on `@HeadElementBuilder`), which forces us to copy the standard headers and add our customizations at the end:
```swift
Head() {
    MetaTag.utf8
    MetaTag.flexibleViewport

    if page.description.isEmpty == false {
        MetaTag(name: "description", content: page.description)
    }

    if context.site.author.isEmpty == false {
        MetaTag(name: "author", content: context.site.author)
    }

    MetaTag.generator

    Title(page.title)

    MetaLink.standardCSS

    if context.site.syntaxHighlighters.isEmpty == false {
        MetaLink.syntaxHighlightingCSS
    }

    if context.site.builtInIconsEnabled {
        MetaLink.iconCSS
    }

    MetaLink(href: page.url, rel: "canonical")

    if let favicon = context.site.favicon {
        MetaLink(href: favicon, rel: .icon)
    }

    // Additional headers
    MetaLink(href: "/css/custom.css", rel: "stylesheet")
[...]

PR #48, opened 2024-05-09
Allow enhancing standard header elements

pd95:
As described in #47 a common problem is that the standard header created by `Ignite` has to be enhanced with some additional elements.

With this PR I propose the enhancement of the current initializer `init(page:,in:)` to take an additional parameter which allows appending more elements to `Head`:

```swift
init(for page: Page, in context: PublishingContext, @HeadElementBuilder additionalItems: () -> [HeadElement] = {[]})
```

This allows enhancing the standard header with as shown below:

```swift
Head(for: page, in: context) {
    // Additional headers
    MetaLink(href: "/css/custom.css", rel: "stylesheet")
    Script(file: "/js/custom.js")
}
```

### DIFF
--- a/Sources/Ignite/Elements/Head.swift
+++ b/Sources/Ignite/Elements/Head.swift
@@ -30,7 +30,7 @@ public struct Head: HTMLRootElement {
     ///   - page: The `Page` you want to create headers for.
     ///   - context: The active `PublishingContext`, which includes
     ///   information about the site being rendered and more.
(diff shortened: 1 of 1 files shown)


pd95:
Additional to above description, I've moved the creation of the "standard set of headers" to `Head.standardHeaders(for:, in:)` for easier re-use. Additionally it will be easier to write tests if we don't have to duplicate the elements.

twostraws:
Very nice – thank you!

PR #49, opened 2024-05-13
Feature: highlight active page in `NavigationBar`

pd95:
This Pull Request implements highlighting the currently active page in the `NavigationBar` as illustrated with the image below. It supports `Link`s and `DropDown`s. The implementation relies on [Bootstraps "active" CSS class for "nav-link" elements](https://getbootstrap.com/docs/5.3/components/navbar/#nav).

Current issues: 

- ~~According to the documentation (see link above) one should add also "aria-current: true" attribute to the currently selected element. **To be implemented**~~
- "On-by-Default": the feature cannot be disabled/enabled. There are no configuration flags.
   **To be discussed**: Do we need configuration for this feature? If yes: wouldn't most users expect this feature to be active by default? This would cause regressions for existing users.
- The current implementation works, but is probably not nice: I had to add a new property `currentRenderingPath` to `PublishingContext` and set it accordingly in some `render` methods to allow checking for it in the `NavigationBar` implementation so we know which link should be marked as "active". I think this would be something to be added to an `@Environment`.

Hightlight-active-page

### DIFF
--- a/Sources/Ignite/Elements/Dropdown.swift
+++ b/Sources/Ignite/Elements/Dropdown.swift
@@ -107,6 +107,7 @@ public struct Dropdown: BlockElement, NavigationItem {
                         if let link = item as? Link {
                             item.class("dropdown-item")
                                 .class(context.currentRenderingPath == link.url ? "active" : nil)
+                                .aria("current", context.currentRenderingPath == link.url ? "page" : nil)
                         } else {
                             item.class("dropdown-header")
                         }
--- a/Sources/Ignite/Elements/Dropdown.swift
+++ b/Sources/Ignite/Elements/Dropdown.swift
@@ -85,9 +85,12 @@ public struct Dropdown: BlockElement, NavigationItem {
     public func render(context: PublishingContext) -> String {
         Group(isTransparent: isNavigationItem) {
             if isNavigationItem {
+                let hasActiveItem = items.reduce(false) { partialResult, item in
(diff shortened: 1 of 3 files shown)


twostraws:
This was something I had hoped to get into the initial release, but it just didn't make the cut because I ran out of time, so I have no problem merging it in now – I don't think it will cause any problems. You've also solved it exactly the right way given the current code structure, so I think we should get this merged in. Are you happy to move forward?

Sources/Ignite/Elements/Dropdown.swift
```diff
@@ -85,9 +85,12 @@ public struct Dropdown: BlockElement, NavigationItem {
     public func render(context: PublishingContext) -> String {
         Group(isTransparent: isNavigationItem) {
             if isNavigationItem {
+                let hasActiveItem = items.reduce(false) { partialResult, item in
+                    partialResult || context.currentRenderingPath == (item as? Link)?.url
+                }
```
PetrPrazak, line 90: No need to iterate all items.
```suggestion
                let hasActiveItem = items.contains { context.currentRenderingPath == ($0 as? Link)?.url  }
```

pd95: Yes that's a nicer solution! Thanks.

twostraws:
Thank you, @PetrPrazak and @pd95! 🙌   All good to merge now?

pd95:
👍🏻 from my side.

PR #54, opened 2024-05-19
Add support for alternative modern image formats

u7k:
- Introduced an initializer for the `Image` struct to accept alternative formats.
- Implemented a method to render <source> elements for known and unknown formats.
- Sorted formats by predefined priority order (heic, avif, jxl, webp).
- Ensured that the `<picture>` element includes all relevant `<source>` elements and a fallback `` element.

### DIFF
--- a/Sources/Ignite/Elements/Image.swift
+++ b/Sources/Ignite/Elements/Image.swift
@@ -142,7 +142,7 @@ public struct Image: BlockElement, InlineElement, LazyLoadable {
         // dictionary where the key is the file extension of the each image.
         let availableFormats = alternativeFormats.reduce(into: [String: String]()) { dict, format in
             guard let lastDotIndex = format.lastIndex(of: ".") else { return }
-            let fileExtension = String(format[format.index(after: lastDotIndex)...])
+            let fileExtension = String(format[format.index(after: lastDotIndex)...]).lowercased()
             dict[fileExtension] = format
         }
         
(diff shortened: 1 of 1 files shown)


twostraws:
Thank you; this is an important improvement.

Given that HEIC and JXL are unsupported by the majority of browsers, and WebP and AVIF are supported by every browser, I'm wondering whether a bigger target here is to take into account retina / dark mode alternative images? Before the initial release, I did some thinking around this, and wondered whether we could verify that image/whatever.png exists at build time and issue a warning if not, and also whether we could detect alternative forms of images using a fixed format, and automatically generate the `srcset` attribute to include them.

For example, if only image/whatever.png exists then it's a simple ``. But if image/whatever@2x.png also exists then we can generate a retina-enabled `<picture>`, if image/whatever~dark.png exists then we could generate a dark mode version as well, and if image/whatever@2x~dark.png exists then we generate retina dark mode. (I further extended this to think that if only @2x exists at build time then it could automatically be downsized to 1x, but that would require a Core Graphics dependency or similar, so I'd rather avoid this for now.)

What do you think?

u7k:
I haven’t done dark mode optimization for the web before, so I can’t provide detailed insights on that aspect. However, it’s crucial to cover the basics first. For instance, images might be stored on a separate service like Google Photos, an S3 bucket, or others. In the image initializer, if the name property isn’t a link (doesn’t start with https://, etc.), then it should prepend the `context.site.url.path` to the relative path and check for file existence using the file manager. If the file doesn’t exist, a warning should be issued.

twostraws:
Yes, that sounds reasonable: if it's a local image then we can try adding retina/dark mode support, but if it's a remote image leave it untouched.

twostraws:
This hasn't gone anywhere in quite a while, and Ignite has changed significantly in the meantime. As a result, I'm going to close this issue.

PR #55, opened 2024-05-19
Add Media Element

dl-alexandre:


twostraws:
Hello! Thank you for this contribution. You haven't added any explanation to this pull request, so it's rather hard to review. Why do we have both `name` and `names`, and both `fileType` and `fileTypes`? Why are file types being specified as a string? Why are there force unwraps scattered through the code? Why did you choose one single `Media` element rather than having `Video` or `Audio` elements?

I realize these are things you've already thought through, but it's important you share your thinking with the rest of us so we can help merge it all. Thank you! 🙂

dl-alexandre:
Yes. Thank you for the questions Paul! I wouldn't have been able to contribute without the help of all your prior work and descriptive [video](https://www.youtube.com/watch?v=JJGLoo3A2uk).

Why do we have both name and names, and both fileType and fileTypes? **Alternative audio files are accepted through players where if one format isn't accepted the browser will try the next. I understand this may be better documented or more intuitively presented?**

Why are file types being specified as a string? **This could possibly become an enum, though I have been confusion by proper format in past cases. (e.g. 'audio/mpeg' or 'audio/m4a' for M4A)**

Why did you choose one single Media element rather than having Video or Audio elements? **The inputs are identical for both with exception of file type. Controls may be similar as well, though my limited undstanding of 'attributes' may have hindered a better design as you can see 'controls' was hardcoded into the element's strings**

twostraws:
> Why do we have both name and names, and both fileType and fileTypes? Alternative audio files are accepted through players where if one format isn't accepted the browser will try the next. I understand this may be better documented or more intuitively presented?

I think perhaps we should use one, non-optional array to store these values. We could then create two initializers, one for a single value and one for an array of values, or look at some other alternatives mentioned below.


> Why are file types being specified as a string? This could possibly become an enum, though I have been confusion by proper format in past cases. (e.g. 'audio/mpeg' or 'audio/m4a' for M4A)

My concern is that we're putting the onus on developers to get the type string right every time, and stringly typed values are easy to screw up. In contrast, we could create enums with string associated values such as these:

```swift
enum VideoType: String {
    case h264 = "video/mp4"
    case h265 = "video/hevc"
    case ogg = "video/ogg"
    case quicktime = "video/quicktime"
    case webm = "video/webm"
}

enum AudioType: String {
    case m4a = "audio/mp4"
    case mp3 = "audio/mpeg"
    case ogg = "audio/ogg"
    case wav = "audio/wav"
    case webm = "audio/webm"
}
```

(Those are just suggestions; I don't know enough about nomenclature here to really nail the enum case names!)

> Why did you choose one single Media element rather than having Video or Audio elements? The inputs are identical for both with exception of file type. Controls may be similar as well, though my limited undstanding of 'attributes' may have hindered a better design as you can see 'controls' was hardcoded into the element's strings.

The inputs are similar, just like how the inputs for `<em>` and `<strong>` are similar, but I think for the sake of clarity at the call site I would feel happier splitting this in two. It would also result in clearer API to support things like enabling/disabling Picture in Picture, `poster`, etc.

I'm still not sure on what would particularly clean API here. We need to know the video (or audio) filename and also its type, which could be rather cumbersome when listing several, and I am keen to avoid separating filenames from their file types, because that's another easy mistake to make. Some options that come to mind to smooth things over:

1. We could create a custom initializer for a `VideoFile` (or `AudioFile`, but you get the point) type that has an optional `VideoType` and auto-detects based on file extension.
2. We could make `VideoFile` conform to `ExpressibleByStringLiteral`, again using auto-detection based on file extension.
3. We could make a `Video` initializer that accepts a `VideoFile`, and separate convenience initializers that accept just a filename string, or a filename string and a `VideoType`.
4. We could make the `Video` type use a variadic parameter for its files rather than users making an array. 

Any of those would help reduce the clutter at the call site, which would be appreciated.

If we went down the route of 2 and 4, then we'd get syntax like this:

- `Video(files: "/videos/example.mp4")`
- `Video(files: "/videos/example.mp4", "/videos/example2.ogg")`
- `Video(files: "/videos/example.mp4", VideoFile("/videos/weirdextension.meh", type: .webm))`

What do you think?

dl-alexandre:
From a user standpoint having separate elements does make more sense now. Especially when filetypes are detected by name not explicitly, I can see how they'd be easily confused within each other.

I didn't have a plan regarding custom files, though [the list](https://www.sitepoint.com/mime-types-complete-list/) used populating types seemed comprehensive enough for a greater majority of cases.

I appreciate for your guidance and direction.

twostraws:
Thank you! I think this is a significant improvement, and sets us up for great media support in the future.

PR #62, opened 2024-05-24
Add Time Element

jobearrr:
This came from a need to index the posts in my blog.

The `time` tag makes it easier to index data by using the information stored in the `datetime` attribute. The `datetime` value must have a valid format (I'm using `ISO8601DateFormatter` to do that).

For instance, the resulting rendering of
``` swift
Time(dateTime: content.date) {
    content.date.asPostListFormart
}
```
 will look like:
``` html
<time datetime="2024-05-24T14:30:00Z">Fri, 24 May 2024 14:30</time>
```
where `asPostListFormat` is an extension of `Date` which prints the date in a human-readable format.

For more details about the `Time` element, please check the [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time).


gavineadie:
I like time on web sites!  Nothing irritates me more than questioning Google and reading a well written answer that's eight years old and entirely out of date.  That's not what @jobearrr is talking about, i know, but for _Ignite_ to default to dropping a `date written` into the page footer would be a delight.  Maybe I'll work on that next.

twostraws:
Thank you – this is a great improvement!

PR #69, opened 2024-06-01
Reading a Markdown file that causes an 'encoding' error throws a confusing error ..

gavineadie:
Ingite reports the reading of a Markdown file with unknown encoding as a parsing error.  This PR catches the file error early and throws that before the parse is tried.

### DIFF
--- a/Sources/Ignite/Publishing/PublishingError.swift
+++ b/Sources/Ignite/Publishing/PublishingError.swift
@@ -17,6 +17,9 @@ public enum PublishingError: LocalizedError {
     /// Invalid Markdown was found at the specific URL.
     case badMarkdown(URL)
 
+    /// A file cannot be opened (bad encoding, etc).
+    case unopenableFile(String)
+
     /// Publishing attempted to remove the build directory, but failed.
     case failedToRemoveBuildDirectory(URL)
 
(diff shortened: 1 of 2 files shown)


twostraws:
Thank you! I'm not sure that throwing an `NSError` is necessarily a good idea, particularly with typed throws incoming shortly. Could we throw `badMarkdown` or create a new error type for this?

gavineadie:
`badMarkdown` is the error that is thrown today!  My concern is that `badMarkdown` isn't the problem (the Markdown text isn't "bad"), but the rather that the `String` initializer fails while reading the given URL because of bad encoding.  So the error happens before the code that might throw `badMarkdown` is invoked.

I'm anxious not to diverge from your style so, just to check before I amend my PR, I propose to catch the `NSError` and throw a new `PublishingError.unreadableURL` (using the `NSError.localDescription`).  That seems to fit with the philosophy of other lower level failures like `failedToCreateBuildFile`.  Would that be OK?

gavineadie:
Ignite will throw a `PublishingError` with text of the form:
```
Failed to open file: The file “EncodingError.md” couldn’t be opened because the text encoding of its contents can’t be determined..
```

twostraws:
Perfect – thank you!

Issue #74, opened 2024-06-18
Close Button


markstamer:
To dismiss modal dialogs I would like to introduce a close button. It seems it is as easy as adding the case `close` to `Role` as Bootstrap will make a close button consisting of a simple cross from `btn-close`. Would this be an appropriate approach or does this polite the `Role` type in any way? I could also introduce a `CloseButton` type which would lead to some duplication though.

PR #78, opened 2024-06-19
Feature / Close button

markstamer:
This PR adds the option to create close button consisting of a cross by setting the role to `close`. It is related to Issue #74 

You can create a button like this:
```swift
Button({}) {
    SomeAction()
}.role(.close)
```
or 
```swift
Button {
    SomeAction()
}
.role(.close)
.onTap
```


twostraws:
Thank you! It would be helpful to add some documentation to IgniteSamples of how modals work, when you have the time.

PR #79, opened 2024-06-19
Feature / Modal View

markstamer:
This PR adds the option to create modal views. It can be configured to look like a modal dialog or by omitting the header and footer show any content provided in the body. 

### Show / Dismiss

It can be shown via the `ShowModal` action and hidden either by tapping on the backdrop or by using the `DismissModal` action. 

### Options

There are a few modifies available to configure e.g. 
* the size
* the backdrop
* dismissal behavior
* vertical position 
* if the content should be scrollable

### Example
```swift
Button("Tap me") {
    ShowModal(id: "myModalId")
}

Modal(id: "myModalId") {
    "Hello, World!"
}
.size(.large)
```

### DIFF
--- a/Sources/Ignite/Actions/DismissModal.swift
+++ b/Sources/Ignite/Actions/DismissModal.swift
@@ -1,0 +1,31 @@
+//
+// DismissModal.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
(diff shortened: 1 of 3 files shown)


twostraws:
This is incredible – I love it! Thank you for contributing this 🙌

PR #82, opened 2024-06-22
Added .newWindow to open new window however keeping .blank as well

NigelGee:
I keeped the .blank for as you said it a prior art for html to use blank and it might break some code. I do feel that it would be useful as I did not realise and took me awhile to find this, and other web site builder seam to use open in new window for their links


twostraws:
Thank you!

PR #83, opened 2024-06-24
Bugfix / Modals not dismissable via Esc key

markstamer:
This PR introduces the html attribute `tabindex` which is necessary to be set on modals to be dismissible via the `Esc` key. 

While I could have just used a custom attribute, I thought modeling it in a "swifty" manner could be of value for other use cases. 

I choose the name `TabFocus` to more clearly hint on its usage but going for `TabIndex` might be more inline with web development. Curious about other opinions here.

### DIFF
--- a/Sources/Ignite/Elements/Modal.swift
+++ b/Sources/Ignite/Elements/Modal.swift
@@ -161,6 +161,7 @@ public struct Modal: PageElement {
             .class(scrollable ? "modal-dialog-scrollable" : nil)
         }
         .class(animated ? "modal fade" : "modal")
+        .tabFocus(.focusable)
         .id(id)
         .aria("labelledby", "modalLabel")
         .aria("hidden", "true")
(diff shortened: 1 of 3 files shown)


twostraws:
Is this commonly customized by developers? I literally have no idea 😅

twostraws:
I looked into this (very briefly!), and it looks like folks generally recommend against using it because it's easy to screw up and cause problem with screen readers. See <https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/#acceptable-uses-for-tabindex> for the list of "acceptable uses" – surprisingly uncommon! That being said, we can still expose it; unless someone disagrees, I think it's better we add the option and perhaps a documentation comment asking users to be careful. How does that sound?

markstamer:
Sounds good. Something along the line. "Use with caution as it is easy to make mistakes and cause problems with screen readers." or more elaborate?

twostraws:
Sounds good to me!

Harry-KNIGHT (approved):

markstamer:
@twostraws You can find the word of caution here https://github.com/twostraws/Ignite/pull/83/commits/e4dda019267d365e1614c61937e8ae7ed060bfb8

twostraws:
Thank you!

PR #85, opened 2024-06-30
Split Makefile into a build and install target

mikaoj:
So building can be done as non-privileged user and only requiring install to be done as root.


twostraws:
Thank you!

PR #88, opened 2024-07-03
Feature / Hover Link Style

markstamer:
This PR adds a new link style called hover. With this style a link is only underlined when hovering over the text.


twostraws:
Very nice – thank you! This will need to be documented at some point.

PR #90, opened 2024-07-03
Feature / Box Shadow on Hover

markstamer:
This PR adds a show and hide shadow action which can be used to show a box shadow on hovering over an element. This is useful to indicated that it can be clicked, e.g. an image that shows a larger version of itself as a modal. 

To be able to reuse the shadows css representation I introduced a `Shadow` type. I'm not happy about the amount of code this adds but also reuses duplication in the css strings. Let me know what you prefer.

```swift 
Image(decorative: "/images/awesome.jpeg")
        .resizable()
        .frame(width: 150)
        .cornerRadius(6)
        .onHover { isHovering in
            if isHovering {
                ShowBoxShadow()
            } else {
                HideBoxShadow()
            }
        }
        .onClick {
            ShowModal(id: "largeImageModalId")
        }
```

### DIFF
--- a/Sources/Ignite/Actions/HideBoxShadow.swift
+++ b/Sources/Ignite/Actions/HideBoxShadow.swift
@@ -1,0 +1,8 @@
+//
+//  File.swift
+//  
+//
+//  Created by Mark Stamer on 03.07.24.
+//
+
+import Foundation
--- a/Sources/Ignite/Actions/HideBoxShadow.swift
+++ b/Sources/Ignite/Actions/HideBoxShadow.swift
@@ -1,8 +1,21 @@
 //
-//  File.swift
-//  
-//
-//  Created by Mark Stamer on 03.07.24.
+// ShowBoxShadow.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
 //
 
 import Foundation
+
+/// Hides a previously applied box shadow of the element it is applied to, e.g. onHover(false).
+public struct HideBoxShadow: Action {
+
+    /// Creates a new HideBoxShadow action.
+    public init() { }
+
+    /// Renders this action using publishing context passed in.
+    /// - Returns: The JavaScript for this action.
+    public func compile() -> String {
+        "this.style.boxShadow=''"
+    }
+}
--- a/Sources/Ignite/Actions/ShowBoxShadow.swift
+++ b/Sources/Ignite/Actions/ShowBoxShadow.swift
@@ -1,0 +1,8 @@
+//
+//  File.swift
+//  
+//
+//  Created by Mark Stamer on 03.07.24.
+//
+
+import Foundation
--- a/Sources/Ignite/Actions/ShowBoxShadow.swift
+++ b/Sources/Ignite/Actions/ShowBoxShadow.swift
@@ -1,8 +1,37 @@
 //
-//  File.swift
-//  
-//
-//  Created by Mark Stamer on 03.07.24.
+// ShowBoxShadow.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
 //
 
 import Foundation
+
+/// Shows a box shadow around the element it is applied e.g. onHover(true)
+public struct ShowBoxShadow: Action {
+    var color: Color = .black.opacity(0.1)
+    var radius: Int = 8
+    var x: Int = 0
+    var y: Int = 0
+    
+    /// Creates a ShowBoxShadow action with the provided shadow parameters
+    /// - Parameters:
+    ///   - color: The shadow's color.
+    ///   - radius: The shadow's radius
+    ///   - x: The X offset for the shadow, specified in pixels. Defaults to 0.
+    ///   - y: The Y offset for the shadow, specified in pixels. Defaults to 0.
+    public init(color: Color = .black.opacity(0.1), radius: Int = 8, x: Int = 0, y: Int = 0) {
+        self.color = color
+        self.radius = radius
+        self.x = x
+        self.y = y
+    }
+
+    var shadow: Shadow {
+        Shadow(color: color, radius: radius, x: x, y: y)
+    }
+
+    public func compile() -> String {
+        "this.style.boxShadow='\(shadow.description)'"
+    }
+}
--- a/Sources/Ignite/Modifiers/HoverEffect.swift
+++ b/Sources/Ignite/Modifiers/HoverEffect.swift
@@ -1,0 +1,85 @@
+//
+// HoverEffect.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+import Foundation
+
+public extension PageElement {
+    /// Applies a hover effect to the page element
+    /// - Parameter effect: A closure that returns the effect to be applied. The argument acts as a placeholder representing this page element.
+    /// - Returns: The page element with the provided hover effect applied.
+    @discardableResult
+    func hoverEffect(_ effect: @escaping (EmptyHoverEffect) -> some HoverEffect) -> Self {
+        onHover { isHovering in
+            if isHovering {
+                ApplyHoverEffects(styles: effect(EmptyHoverEffect()).styles)
+            } else {
+                RemoveHoverEffects()
+            }
+        }
+    }
+}
+
+/// A protocol representing the hover effect css styles to be applied
+public protocol HoverEffect {
+    var styles: [AttributeValue] { get set }
+}
+
+/// An empty hover effect type to which styles can be added
+public struct EmptyHoverEffect: HoverEffect {
+    public var styles: [AttributeValue] = []
+}
+
+extension HoverEffect {
+    /// Adds a shadow hover effect
+    /// - Parameters:
+    ///   - color: The shadow's color. Defaults to black at 33% opacity.
+    ///   - radius: The shadow's radius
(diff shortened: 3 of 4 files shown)


twostraws:
I like the concept (you're on fire!), but I'm not sure about the execution. If we take this approach we'll end up with a  bunch of actions that are really just modifiers being conditionally applied. If we're going to do support on hover variants of our elements, we might want to look for a more all-encompassing approach that lets us unlock other kinds of modifiers too.

For example – and this is literally off the top of my head, so please don't take this as me saying this is the Right Way™ – what if we were to take inspiration from SwiftUI's `visualEffect()` modifier to make some kind of `hoverEffect()` modifier? Visual effects pass the view that they are modifying into a closure, at which point a bunch of different modifiers can be applied. So, using your example we might have this:

```swift
Image(decorative: "/images/awesome.jpeg")
        .resizable()
        .frame(width: 150)
        .cornerRadius(6)
        .hoverEffect { element in
                element
                        .shadow(radius: 5)
        }
        .onClick {
            ShowModal(id: "largeImageModalId")
        }
```

SwiftUI only supports a subset of modifiers inside visual effects, and we could do the same – start small, perhaps just with shadows, and work our way up. I think even though they look like regular modifiers, internally they would become JavaScript code to be executed when the hover state changes.

We'd need to think carefully about reverting all the style changes when hovering ends, but I think we could get that fairly easily. For example, the below sample HTML applies a bunch of styles to some text, stashes those styles away in a local property of the element when hovering starts, then restores them when hovering ends:

```html
<html>
<body>
<p id="test" style="background-color: blue; padding: 50px; border-radius: 10px;" onmouseover="this.unhoveredStyle = this.style.cssText; this.style.backgroundColor = 'red'; this.style.borderRadius = '50px';" onmouseout="this.style.cssText = this.unhoveredStyle;">This is a test.</p>
</body>
</html>
```

Presumably we'd also need to do something for classes, if/when we support for that inside hover effects.

markstamer:
That is indeed a nice API.  Let me look into it. 

> (you're on fire!)
I'm building my portfolio website at the moment and these PRs are basically things I build to achieve the look I want and I'm happy to share.

markstamer:
I have checked the SwiftUI APIs for the new `HoverEffect`, which is in beta, and the `VisualEffect` you suggested. Both would work but their approach is different.

### Option I
SwiftUIs new `HoverEffect` takes a semantic approach. It's a simple struct and there are static properties defined on it called `lift` and `highlight`. However, you can create a `CustomHoverEffect` if needed. To my understanding in our case a hover effect is just a bunch of css styles so a custom effect could be as simple as exposing a initializer like this `HoverEffect(styles: [AttributeValue])` or limited to existing effects `static func combine(_ effects: [HoverEffect]) -> Self`.

Even though choosing a set of effects that work for most people is a little tricky this looks like a simple but flexible approach to me.

```swift 
 element
    .hoverEffect(.lift(.level1))

 element
    .hoverEffect(.myHoverEffect)
``` 

### Option II 
In contrast `VisualEffect` is a protocol and effects are applied via methods on `VisualEffect` to the `EmptyVisualEffect` which is passed in the closure. It looks like the effects target more or less a single style property. I guess this approach would be less opinionated of what a good hover effect is but needs more tinkering.

My approach to this implementation would be to pass the `EmptyHoverEffect` through the chain off effect methods and add the styles to an array in each method.

```swift
protocol HoverEffect {
    var styles: [AttributeValue] { get  set }
}

struct EmptyHoverEffect: HoverEffect {
    var styles: [AttributeValue] = []
}

extension HoverEffect {
    func shadow() -> some HoverEffect {
        addingStyle(.init(name: "boxShadow", value: "shadow values"))
    }

    func background(_ color: Color) -> some HoverEffect {
        addingStyle(.init(name: "backgroundColor", value: "\(color)"))
    }

    func addingStyle(_ style: AttributeValue) -> Self {
        var copy = self
        copy.styles.append(style)
        return copy
    }
}

extension PageElement {
    @discardableResult
    func hoverEffect(_ effect: @escaping (EmptyHoverEffect) -> some HoverEffect) -> Self {
        let styles = effect(EmptyHoverEffect()).styles
        onHover { isHovering in
            // Use styles here
        }
        return self
    }
}

// Usage

element
    .hoverEffect { content in
        content
            .shadow()
            .background(.blue)
    }
```

### Reverting Styles
> onmouseover="this.unhoveredStyle = this.style.cssText;

This was a great hint but it doesn't work e.g. for `boxShadow` because if there is no shadow applied in the first place it is later not removed. However, I can save the styles explicitly like this: `"this.dataset.original\($0.name.capitalized) = this.style\`($0.name)`"`

twostraws:
Hello! First, let me apologize again for taking so long to reply – moving house has been hard! However, I'm back on the case now.

> This was a great hint but it doesn't work e.g. for boxShadow because if there is no shadow applied in the first place it is later not removed.

That's interesting! I tried it with a box shadow and it seemed to work. What am I missing?

My code:

```html
<html>
<body>

<p id="test" style="background-color: blue; padding: 50px; border-radius: 10px;" onmouseover="this.unhoveredStyle = this.style.cssText; this.style.backgroundColor = 'red'; this.style.borderRadius = '50px'; this.style.boxShadow = '10px 5px 50px red';" onmouseout="this.style.cssText = this.unhoveredStyle;">This is a test.</p>

</body>
</html>
```

markstamer:
> Hello! First, let me apologize again for taking so long to reply – moving house has been hard! However, I'm back on the case now.

I can imagine. Hope you are set and well after all. Glad you are back!

> That's interesting! I tried it with a box shadow and it seemed to work. What am I missing?

I tired it again and found that it works indeed. 🙈 I can only assume I made a typo in the remove action and couldn't think of any other explanation.

> Also: to be clear, I was suggesting to take SwiftUI's visualEffect() modifier as inspiration, but name it hoverEffect() in Ignite.

 Let me know what you think about my last commit. I got from you comment that you prefer option II.

twostraws:
I'm going to merge this in and take a look more closely; it still doesn't quite land where I'm hoping for, but it's already a big step forward. Thank you!

twostraws:
@markstamer Okay, I've made a handful of changes that I think gets it closer to the (admittedly invisible!) goal I had in my head. Could you please cast an eye over the "hover-achiever" branch and see what you think? What I've done is push `HoverEffect` way up the tree, so most elements to that protocol.

This allows us to re-use a wide variety of effects from other modifiers - things like `foregroundStyle()`, `opacity()`, `cornerRadius()`, `textDecoration()`, etc, now all work as hover effect modifiers, for example. My goal is to reduce duplication and widen what can be achieved, and this kind of code works great:

```swift
Text("Welcome to ExampleSite!")
    .font(.title1)
    .hoverEffect { content in
        content
            .foregroundStyle(.white)
            .background(.red)
            .shadow(radius: 3)
            .opacity(0.5)
            .cornerRadius("25%")
            .textDecoration(.underline)
    }
```swift

If that all looks sensible to you, we can rationalize things a little – `HoverEffect` is not a great protocol name with this new approach, I doubt I was comprehensive in my documentation, and I think some code is in the wrong files; I was mostly noodling around. I'm imagining `PageElement` being used for things that aren't hover compatible – changing the accessibility label, adding a hover effect (a hovering hover effect?), changing the hint, etc, all don't really make sense as hover effects.

If you're happy with these changes, I'll merge them into main.

markstamer:
@twostraws I took a look at your branch and it definitely makes sense to bundle the style related modifiers under a single protocol. Keeping the name `HoverEffect`, however, is confusing. What about changing `HoverEffect` to be a style focused protocol, such as `CSSStylable`? All style modifiers would be on this protocol and the `HoverEffect` protocol conforms to it to keep the API we started with. 

What still bugs me is that you could add events and classes to it. This is a bit more invasive but I guess there is no clean way unless the `CoreAttributes` type is split in different types, since it mingles id, styles, classes, aria, etc. at the moment. This is only an idea at the moment but what if we split it in CoreAttributes and StyleAttributes and create separate protocols to hold them. So what if 

* `BaseElement` becomes `HTMLRenderable` 
* The new `BaseElement` protocol gets the `CoreAttributes` type but without styles
* `CSSStylable` gets a `StyleAttributes` type and all modifiers that address styling

This way 
* `Component` conforms to `HTMLRenderable` instead of `BaseElement`
* `PageElement` conforms to `HTMLRenderable, `BaseElement` and `CSSStylable` 
* `HoverEffect` conforms to `CSSStylable` but *not* to `BaseElement` 

Of course we would need to go to all types conforming to `PageElement` and add a `StyleAttributes` property but it would solve the issue, I guess.

I can totally understand if at his point you don't want to mess with the base protocols anymore and be pragmatic but I think it would follow the path of your original intention until the end, no?

twostraws:
I'm happy to mess with the base protocols – we're still not near a 1.0 😂 

I merged in a fairly wide-ranging PR from @dl-alexandre, so I'll need to probably do bits of this work again. I'm on the case right now, and should have it merged soon.

markstamer:
Great! I tried it out quick and dirty to see if there are any blockers but it seems work. Adding the styles property to all page elements is somewhat annoying though.

twostraws:
> Adding the styles property to all page elements is somewhat annoying though.

Can you give me an example of what you mean?

twostraws:
I've merged these changes in now, and ended up with a much simpler approach. I was thinking previously that keeping some kind of hover distinction in place would be good for protocols, but then I kind of figured if someone wants to do something "creative" – add a hover effect to a hover effect – they are welcome to try! It certainly makes the code much easier, because it just moved things from `HoverEffect` to `PageElement`.

PR #93, opened 2024-07-08
Feature / Background Image Modifier for Page Elements

markstamer:
This PR adds a background image modifier for page elements. The basic usage is:

```swift
element
    .backgroundImage("/images/image-name.png", contentMode: .fill)
```

The default position is centered. To adjust the position of the background image there is an additional `.backgroundImagePosition()` modifier.

### DIFF
--- a/Sources/Ignite/Modifiers/BackgroundImage.swift
+++ b/Sources/Ignite/Modifiers/BackgroundImage.swift
@@ -199,7 +199,8 @@ public struct BackgroundPosition: CSSRepresentable {
 
 private extension PageElement {
     func style(_ value: String, replace: Bool) -> Self {
-        if let existingIndex = attributes.styles.firstIndex(where: { $0.hasPrefix(value) }) {
+        let key = String(value.prefix(while: { $0 != ":" }))
+        if let existingIndex = attributes.styles.firstIndex(where: { $0.hasPrefix(key) }) {
             guard replace == true else {
                 return self
             }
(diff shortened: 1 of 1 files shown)


twostraws:
This looks great – thank you! I'm thinking about how this fits in with the current `background()` modifier, which was previously called `backgroundColor()`. Seeing this PR, `backgroundColor()` would be a really nice complement to `backgroundImage()`, and I certainly prefer uniformity if possible.

Possible compromise: `background(color:)`, `background(image:)`, and `background(styles:)` – that feels nice and neat to me. What do you think?

Regarding `backgroundImagePosition()`, could this brought in as a parameter to the existing `backgroundImage()`/`background(image:)` modifier?

markstamer:
I'm fine with either naming convention. I can change it to `background(image:)`. I think I just missed that you deprecated `backgroundColor()`.

I started out to have the position as a parameter but found the function got really clunky at some point and moved it out. I can make it a parameter again. I think it even simplifies the logic I need to add to account for different modifier orders.

twostraws:
Thank you!

Sources/Ignite/Modifiers/BackgroundImage.swift
```diff
@@ -0,0 +1,201 @@
...
+}
+
+private protocol CSSRepresentable {
+    var css: String { get }
+}
+
+/// The possible background sizes
+public enum BackgroundImageContentMode: CSSRepresentable {
+    /// This is the default value. The background image is displayed at its original size.
+    case original
```
twostraws, line 29: Should this be the default value for the backgroundImage() method, then? It would mean folks can just call background(image: "………") and get sensible defaults.

PR #100, opened 2024-07-31
Update Embed.swift with Embedded Map

dl-alexandre:
Embedded Maps are simple enough to use Embed Element whereas more complex maps will have their own element with differing parameters.


twostraws:
Let's focus on getting the `Map` working well. Once that API lands, I think a convenience initializer for `Map` would work better than `Embed`, to keep things nice and clear.

Issue #86, opened 2024-07-02
Support custom install locations in Makefile


compufox:
Currently the Makefile is hardcoded to only install to `/usr/local/bin` which is not always possible or desired.

It would be nice to have location controlled via shell variable or some other command line invocation

PR #102, opened 2024-07-31
allow install directory in makefile to be overridden

compufox:


compufox:
addresses #86

twostraws:
Thank you! I'll merge this in, but could you add some documentation for it, please?

PR #107, opened 2024-08-07
Make Bootstrap files optional

0xTim:
Many sites may want control over the bootstrap files that are added, for example:

* Using a pipeline where the CSS is generated by SASS and JS with NPM
* Using GH pages where we need to use the files from the Bootstrap CDN
* Using a different design framework altogether

This PR introduces a new config option, `useDefaultBootstrapURLS`, that can be used to disable the default CSS and JS files for Bootstrap to provide better customisation.

### DIFF
--- a/Sources/Ignite/Components/FeedLink.swift
+++ b/Sources/Ignite/Components/FeedLink.swift
@@ -14,7 +14,7 @@ struct FeedLink: Component {
     func body(context: PublishingContext) -> [any PageElement] {
         if context.site.isFeedEnabled {
             Text {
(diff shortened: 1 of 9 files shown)


0xTim:
@mikaelacaron if/when this is merged you should pull this in and disable it to stop it conflicting with the CDN files

0xTim:
@twostraws yep sounds good! I've updated the PR

Note that I've also applied the same options for the Icon stuff, which should probably be a separate PR, but ties in. This would be a breaking change though, I don't know how much of a concern that is

PR #108, opened 2024-08-14
Add Progressive Web App Manifest File

dl-alexandre:
This addition of a manifest file makes sites as a standalone app when linked on (iOS) home screen rather than a link which opens safari. Mac supports [this function](https://support.apple.com/en-us/104996) automatically.


twostraws:
I like the idea, but doesn't this generate the same manifest file for all sites? That's unlikely to be welcome 😅

dl-alexandre:
Here options were applied setting default implementation as NOT creating a web app by default. This still allows advanced Ignite use cases with Web App functionality. Having a `manifest.json` file generated by site parameters is in the works.

Issue #112, opened 2024-08-29
Feature Cursor change on hover


markstamer:
### What is missing
In case of using `onClick` instead of a link or button the cursor does not change to a hand to indicate that something is clickable. 

### How we could add it
There seems to be a simple style one could use `.style("cursor: pointer")` to archive this. I can imagine a `Cursor` type and to directly use this as a parameter for the onClick method or just as a modifier like `func cursor(_ cursor: Cursor) -> PageElement`. 

Any thoughts?

twostraws:
I think `func cursor(_ cursor: Cursor) -> PageElement` is exactly correct 👍

markstamer:
All right, will make a PR at times.

PR #121, opened 2024-09-17
Feature / Cursor Styles for Page Elements

markstamer:
This PR adds a couple of common cursor styles and a modifier on PageElement to apply them.

See https://github.com/twostraws/Ignite/issues/112 for reference.


twostraws:
Thank you!

PR #122, opened 2024-09-17
Added table with links to website and repository created with Ignite

jcalderita:
This change includes a link to a repository that only contains a README with a table of websites made with Ignite and its repositories.

I believe that the repositories shared here should share the Ignite project policy and be OpenSource.

Another repository has been created for as discussed in another pull request, so that the current one is not contaminated with many pull requests.

The ownership of this new repository should still be @twostraws 

Whatever is decided.


jcalderita:
Finally it will go into the original repository.

PR #125, opened 2024-09-21
Update README.md: adds instructions to add a layout

appfrosch:
Adding a layout is required to enable rendering of markdown files. Instructions for that had been missing so far.


twostraws:
Thank you!

Issue #126, opened 2024-09-21
Copying assets from wrong directory


vdhamer:
When Ignite is added to an app using "add package", it stores the generated site in the Build subdirectory in the app's sandbox directory. This works since PR #119.

But it also needs to fetch Assets/Content/Includes files from the app's sandbox directory, because a sandboxed app cannot access any directory - even for read-only access. Attempting to access the source code subdirectories results in errors about not enough privileges (even if all the Unix level access rights are sufficient).

[Reference](https://www.appcoda.com/mac-app-sandbox/) on file access from a sandbox (highly readable).

I will submit a new PR.

Issue #81, opened 2024-06-22
Swift 6 build fails


multitudes:
I thought this is interesting. It doesn't find the `URL` extension when compiling with the Swift 6 keychain.
The output is:
```
Building for debugging...
[0/5] Write sources
[1/5] Write swift-version-6A63A117629E05AE.txt
error: emit-module command failed with exit code 1 (use -v to see invocation)
[3/7] Emitting module IgniteStarter
/Users/laurentb/Developer/Ignite/ExampleSite/Sources/Site.swift:20:15: error: ambiguous use of 'init(_:)'
18 |     var name = "Hello World"
19 |     var titleSuffix = "My Awesome Site"
20 |     var url = URL("https://www.example.com")
   |               `- error: ambiguous use of 'init(_:)'
21 |     var builtInIconsEnabled = true
```

Swift 5.10 has no problem with it!

twostraws:
It's not ideal, but we could change this. I'm really curious what's causing the breakage here!

clarkezone:
I'm hitting this issue also

Rspoon3:
Still an issue with Version 16.0 beta 6 (16A5230g).

ppporch:
Xcode Version 16.0 beta 6 (16A5230g) has an issue with var url: URL = URL(static: "https://www.example.com"). I get an error message of extraneous argument label 'static:' in call. To fix, remove static:

URL = URL("https://www.example.com") works

appfrosch:
> This should work now in Swift 6, tested with `swift:6.0`

You mean without having to change anything? If so, how–cause I just started looking into Ignite yesterday and had this error as well without any code changes after creating a new site with `ignite new MeWebsite`, using the latest Xcode Version 16.0 (16A242d) …

I also tried to change the Package to this without success:
```swift
// swift-tools-version: 6.0
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "IgniteStarter",
    platforms: [.macOS(.v13)],
    dependencies: [
        .package(url: "https://github.com/twostraws/Ignite.git", branch: "main")
    ],
    targets: [
        .executableTarget(
            name: "IgniteStarter",
            dependencies: ["Ignite"]),
    ],
    swiftLanguageModes: [.v6]
)

```

I got around this by changing the line to `var url: URL = URL("https://www.example.com")` (as stated in this issue thread).

JPToroDev:
[...]

PR #128, opened 2024-09-23
Copying assets from the wrong directory

vdhamer:
Contains 3 changes:

* Fix: #126 
This is allows apps that include Ignite via "add package" to access Assets/Content/Includes data.
Related to MacOS sandboxing.
* added a row to the table with websites that use Ignite ( @jcalderita )
* 2 tests gave `Ambiguous use of 'init(_:)` errors under Xcode 16.0 and Swift 6
Had to add `: URL` as a workaround. Not sure what the root cause is. See #81


twostraws:
Thank you!

PR #131, opened 2024-10-03
Fix image duplication in Card

beny:
Image was currently rendered first and also after items before footer. This will revert back the change brought by last code update.


twostraws:
Thank you!

PR #132, opened 2024-10-09
Compliment XCTests with Swift Testing

dl-alexandre:
There was trouble in running the tests, something with the schemes on my end. Only could run on scheme ignite-package, though they'd crash.


dl-alexandre:
Also PublishingContext required changes because of missing `from:` argument, plus extra `root` argument

twostraws:
Hello! Thank you for this contribution. Why are we moving to Swift Testing?

dl-alexandre:
A Swift Testing move isn't the wrong idea. A couple tests in this repo were some of my earliest testing attempts using XCTest. Learning Swift Testing has been more intuitive than working with XCTest. Plus, providing functional examples. For me, functional examples are the primer for understanding or reference new practices better. Given that this repo is still in development phases, implementing Swift Testing as its standard isn't a heavy lift. Hopefully this clarifies this, even if somewhat opinionated.

twostraws:
I understand you might find it easier to learn Swift Testing, but there are much better ways to learn it than adopting it in a public project, and it would also cut us off from the many thousands of people who already use XCTest. While I could understand moving to Swift Testing once it has gained more traction, right now it has a really tiny percentage share of usage in the Swift world, so I'd rather stay away from it here for the time being.

dl-alexandre:
Here's a split testing commit, let me know if you'd like that changed.

twostraws:
Unless I'm reading it wrong, this still seems to be bringing Swift Testing into the project, which again is not something I would support at this time.

twostraws:
We're now a couple of months on – is there much evidence that other projects are moving to Swift Testing? I did see [one project](https://github.com/johnfairh/SourceMapper) make the move, but I'm reticent to jump ship unless we have a compelling reason.

twostraws:
I'm keen to either move ahead with this, or close it; I don't want to leave it lingering.

dl-alexandre:
Thank you for your patience regarding this.
- Two Testing Targets are present and both run when called
    - Swift Testing -> IgniteTesting
    - XCTests -> IgniteTests
- All tests work but not all test may be helpful
   - **NavigationBar** Test needs some work as does the Element itself as previous test weren't building
   - **Text** Test had some inaccuracies. Need confirmation that test change reflects accurate use specifically `test_customFont()` Note: Swift Testing will allow this test to be rewritten using arguments at some point

Tests/IgniteTests/IgniteTests.swift
```diff
@@ -5,16 +5,84 @@
...
     /// A publishing context with sample values for root site tests.
-    let publishingContext = try! PublishingContext(for: TestSite(), from: #filePath)
+    let publishingContext = try! PublishingContext(
+        for: TestSite(), from: #filePath)
     /// A publishing context with sample values for subsite tests.
-    let publishingSubsiteContext = try! PublishingContext(for: TestSubsite(), from: #filePath)
+    let publishingSubsiteContext = try! PublishingContext(
+        for: TestSubsite(), from: #filePath)
+
+    public func normalizeHTML(_ html: String) -> String {
```
dl-alexandre, line 22: Tests fail at random otherwise

dl-alexandre (commented):
Some HTML data is normalized for easier comparison

twostraws:
Are you happy for this to be merged? It would mean pushing ahead with Swift Testing at a fairly fast pace.

dl-alexandre:
Yessir. There's not much complexity within the test just yet, we still have XCTests as comparisons, and after some more familiarizing I'm confident and comfortable these test will continue to prove useful.

twostraws:
Okay, then let's give it a whirl!

PR #133, opened 2024-10-09
Swift 6

dl-alexandre:
This making a few objects sendable including
- Action
- ShowModal
    - Option
- ContentPosition
- ContentAlignment
- MetaLink
- MetaTag
- Event
- FeedConfiguration
    - ContentMode
    - FeedImage
- UnitPoint
- AttributeValue
- Color
- CoreAttributes
- DiagonalEdge
- Edge

The Version String is also preferably `let` + Updated for current v0.2.1

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 6.0
+// swift-tools-version: 5.9
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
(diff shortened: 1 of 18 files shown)


twostraws:
Thank you – this looks great! Almost all the changes look good, but do we *need* to bump the Swift tools version number? It would restrict Ignite to folks using Swift 6 or later, which is far from everyone at this point.

twostraws:
I'm referring to someone who wants to build Ignite as a standalone project, who doesn't have Swift 6.

dl-alexandre:
Understandable, Package.swift adjustments should be handled by you.

twostraws:
Thank you! Once we actually need some Swift 6 features we can bump the version up. Until then, using an older swift-tools version should allow more people to use Ignite.

PR #134, opened 2024-10-11
Update CoreAttributes.swift

jpurnell:
custom styles were not distinguished from heading, and was failing upon build. Restored space.


twostraws:
Thank you!

PR #137, opened 2024-10-22
added specific versioning for swift-markdown

compufox:
Swift-Markdown recently started cutting versions. I decided to go ahead and update the package definition to use the new versions, allowing Ignite to be built inside of an existing project.

More details here: https://github.com/swiftlang/swift-markdown/issues/178#issuecomment-2177095161


twostraws:
Thank you!

PR #138, opened 2024-10-26
Ensure soft breaks in Markdown result in spaces in the resulting HTML

harlanhaskins:
When Markdown is wrapped to line limits, newlines in the markdown result in a soft break. This break should expand to a single space, and two newlines result in a proper newline.

This fixes the MarkdownToHTML generation to ensure soft breaks are spaced properly.

This also fixes a build error in the tests.


twostraws:
Thank you!

PR #140, opened 2024-11-02
Create CustomFont Interface

froggomad:
### Summary
This PR introduces a new struct, `CustomFont`, and extends `PageElement` to support custom font management, allowing developers to define, add, and use custom fonts easily.

### Key Changes
1. **`CustomFont` Struct**:
   - Provides a way to define a custom font using a font family name and a URL string (which can be local or remote).
   - Adds optional fallback font families that are joined in CSS syntax, allowing for a smooth transition if the primary font fails to load.
   - Includes a computed property, `fontURLString`, which returns a valid URL string or an "Invalid URL" message if the URL is malformed.

2. **`PageElement` Extensions**:
   - `addCustomFontFace(_:)`: Adds a CSS `@font-face` rule using the `CustomFont` instance to make the font available across the application.
   - `useCustomFont(_:)`: Applies the custom font to elements, leveraging the fallback font stack if specified.

### Example Usage
The PR provides an example for usage with an enum, `BrandFont`, that returns predefined `CustomFont` instances. The example demonstrates adding the custom font faces to the HTML and using a specific font as a default on the Body element.

### Additional Notes
- This setup supports local and remote font resources by specifying a valid URL.
- The structured approach ensures that custom fonts are consistently referenced, providing flexibility with fallback options and easy integration into different parts of the project.

### Benefits
- Enhances the styling capabilities by making font integration more straightforward and scalable.
- Improves code readability and reuse by centralizing font management.

### Testing
- Ensure custom fonts load correctly in various browsers and environments (local/remote fonts).
- Verify fallback behavior by disabling network access to test fallback font rendering.

### DIFF
--- a/Sources/Ignite/Styles/CustomFont.swift
+++ b/Sources/Ignite/Styles/CustomFont.swift
@@ -1,0 +1,83 @@
+public struct CustomFont {
+    /// The URL to the font resource (might be local or remote)
+    /// - Returns: String containing a valid URL or "Invalid URL: `urlString`"
+    public var fontURLString: String {
+        fontURL?.absoluteString ?? "Invalid URL: \(urlString)"
+    }
+    /// The font-family name used in the resource (ex .ttf file)
+    public let fontFamily: String
+    /// The URL to the font resource (ex .ttf file)
+    private let urlString: String
+    private var fontURL: URL? {
+        URL(string: urlString)
+    }
+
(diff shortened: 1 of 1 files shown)


twostraws:
Hello! Thank you for this contribution – I think it's an important step forward. I'm going to merge it in now, but I want to find some time to noodle around with the API a little more to see if we can make it a little more transparent in use.

PR #141, opened 2024-11-04
Adding underline opacity style to Link

B4DG3R:
## Information

This PR adds link opacity to the `Link` class for both `underline-opacity-{opacity}` and `underline-opacity-{opacity}-hover`.

The new functionality allows the original `.hover` style functionality to take precedence, where as if the user adds styling to either the underline or hover underline, this then takes presidence.

I'm no web dev so please be brutal 😅, all advice is taken onboard, and I'm not married to the names, if you can think of something better than please suggest one 🙂 👍.

## Changes

- Added a `UnderlineStyle` enum to the link class, allowing selection between `.default`, `.plain` and `.opaacity`.
- Added both an `underline` and `hoverUnderline` variables to the `Link` class, both defaulted to `.default`.
- Added the styling for the new properties inside the `linkClasses` computed property.
- Added two new functions underlineStyle and hoverUnderlineStyle allowing the user to set the new style as and where required.

## Evidence
`underline` opacity 25%
:-: 
opacity 25

`underline` opacity 50%
:-: 
opacity 50

`underline` opacity 75%
:-: 
opacity 75

`underline` opacity 10% `hoverUnderline` opacity 50%
:-:


twostraws:
Thank you for submitting this! You've added an important feature, and one I'm keen to bring in. I've been staring at the code for a while, trying to figure out if there's a way to polish up the API just a little. Specifically, I'm keen to:

1. Avoid situations where people  use `linkStyle(.button)`, then wonder why their underline/hover styles aren't showing.
2. Avoid adding extra modifiers that do similar things to `linkStyle()`.

I also think Bootstrap's API is a little odd here – they've defined a handful of very specific numbers, and my instinct is to say "why can't I have opacity 33%?" etc.

So, I'd like to bat about a few alternatives with you, and perhaps input from some other contributors – @JPToroDev, @vdhamer, and @markstamer would be sure to have some useful insight too, if they have the time.

I apologize ahead of time: getting naming right is tricky, and largely subjective. The below is my opinion, and I'm certainly keen to hear other suggestions.

For example, what if we modified the `LinkStyle` struct to this:

```swift
public enum LinkStyle {
    case `default`
    case hover
    case button
    custom(base: UnderlineStyle = .default, hover: UnderlineStyle = .default)
}
```

That would mean using code like the below:

```swift
Link("Click Me", target: "/")
    .linkStyle(.custom(base: .plain, hover: .opacity50))
```

(Note: Is `base` the right name here? `default` doesn't work, because it would mean the call site can be: `default: .default`! It's telling that Apple uses `.automatic` in place of `.default`, which might help us resolve this neatly. I'm not even sure that `.custom(…)` is the right name here. Maybe `.underline(…)`? That could lead to `.linkStyle(.underline(default: .automatic, hover: .plain))`.

Alternatively, we could still leave something like `custom(base: UnderlineStyle, hover: UnderlineStyle)` in there, but also add a `linkStyle()` overload that avoids the `.custom` clutter – something like `.linkStyle(base: .plain, hover: .opacity50)`, which internally wraps those values in a `custom(base: UnderlineStyle, hover: UnderlineStyle)` case.

(We could perhaps also rename `hover` to something more descriptive, such as `hoverUnderline` or `underlineOnHover`?)

A second alternative is to introduce a `LinkStyle` protocol that creates custom link styles. This doesn't need to be complex, and could be as simple as something like this:

```swift
protocol LinkStyle {
    var baseClasses: String { get }
    var hoverClasses: String { get }
}
```

We could perhaps extend that to allow `baseStyles` and `hoverStyles` as well to give maximum flexibility, although we'd need to provide sensible default values to make it less annoying to use – it would be nice to have to change only the properties you care about.

That protocol could then either be used for all `linkStyle()` values – i.e., `.button` becomes an implementation of the `LinkStyle` protocol, etc, or if that proved tricky then we could use it only for the `.custom` case.

The third thing rattling around my head is the cases inside `UnderlineStyle`. A lot depends on where we land with the other stuff, but:

1. I'm not sure that `plain` is sufficiently different from `default`. Perhaps `none` might work? Or `noUnderline`? If we ended up with a `linkStyle` case called `underline()`, then perhaps this could be `.never`?
2. Are the `link-underline-opacity-*` Bootstrap classes any different to using an inline CSS style? If not, then we could remove them entirely and let users specify the opacity they want.
3. If there _is_ an advantage to using the Bootstrap classes, then I wonder whether we would be better to select the important ones then hide their actual meaning behind some wording? For example, `.opaque` for 100% opacity, `.translucent` for 50% opacity, and `.barelyVisible` for 25% opacity? (This feels like a better idea than number 2, because I really can't imagine users asking for opacity 47% (for example).)

Anyway, these are just my rather chaotic thoughts – the concept is definitely sound, I just want to make sure we get a nice and clean API that keeps clutter low, and also gives users lots of scope to get exactly the result they want. Hopefully some other folks have suggestions!

B4DG3R:
Hey Paul,  thanks for the brain dump 😂!  I'm glad you like the idea but yes you are correct it needs to be right.  I wasn't 100% sure which way to go with it, so after spending some time thinking about it, I came up with this and thought if it was good enough we could discuss it and get it right through PR 🙂.

I am happy to take suggestions to get it right.  I think avoiding people having to add multiple modifiers to get their desired effect is a good way to go, also avoiding modifiers that cancel each other out.

So a single function which can take multiple parameters would take my favor.

I also agree that the Bootstrap opacity values seem bonkers.  Why you can't just add a custom value is beyond me 😅.

One thing I will mention which I am also keen to get working while we are fleshing this out and getting it right, is being able to change the colour of a link and alter the underline and hover properties at the same time.  So currently if I do:

```
                Link(content)
                    .foregroundStyle(Color.aquamarine)
```

Then we can change the link colour, but if we add the `underlineStyle` or `hoverUnderlineStyle`, the `foregroundStyle` modifier gets overridden and the link remains blue (I realised this tonight before writing this).

```
                Link(content)
                    .foregroundStyle(Color.aquamarine)
                    .underlineStyle(.plain)
                    .hoverUnderlineStyle(.plain)
```

I'm not trying to change the direction of the PR, but just wanted to put this here for people to think about if we're taking additional input, as I would hate to add an implementation that then bottlenecks a future feature.

I'll take another look at it tomorrow to see how it can be improved and I'll come back to you with something.  I'm also keen to see what others think about how it can be improved 🙂.

twostraws:
<< I would hate to add an implementation that then bottlenecks a future feature. >>

We're still very firmly in the 0.x version numbers, so I'm happy to break things for the time being – we can always change our mind in a month or two if things didn't quite pan out!

JPToroDev:
Thanks for this great feature, Matt—and for inviting me to the conversation, Paul! Funny how it’s often the more granular APIs that require extra finesse. I think leaning into that granularity with the terminology might be beneficial. Here’s what I might suggest based on what everyone’s said (and interpreting `button` to mean no underline):

```swift
enum LinkDecoration {
    case automatic // Underline with sensible defaults
    case none // No underline
    case underline(_:, hover:) // "inactive" or "base" could also work well
}
```

With usage like this:

```swift
Link("Click Me", target: "/")
    .linkDecoration(.none)
```

I agree that an overload for the `underline` case would be nice!

> I wonder whether we would be better to select the important ones then hide their actual meaning behind some wording… I really can't imagine users asking for opacity 47%

Agreed. I like this idea a lot, and I think it might be a good opportunity to take inspiration from SwiftUI and create an `UnderlineProminence` type, with cases like `none`, `standard`, `increased`, and `decreased`—or cases like `faint`, `light`, `heavy`, etc.

So we'd have something like:

```swift
Link("Click Me", target: "/")
    .linkDecoration(.light, hover: .heavy)
```

I think we can get `foregroundStyle(`) to play nice with this API with some small tweaks to its implementation—I don’t think it throws a wrench into what we’re doing here. 🙌🏼

Let me know what y'all think!

twostraws:
I'm happy with `linkDecoration(.light, hover: .heavy)`. If folks then apply a `linkStyle()` for buttons as well, it feels like user error 😅

Is that okay, @B4DG3R?

B4DG3R:
Works for me 🙂, thanks for the input @JPToroDev.

Being able to specify what you want in a single function like `linkDecoration(.light, hover: .heavy)` seems much more natural.

> If folks then apply a linkStyle() for buttons as well, it feels like user error 😅.

@twostraws, Just so I have this right, do you mean if someone were to apply a `linkDecoration`, on top of a `linkStyle` it would be user error? like below?

```swift
Link("Click Me", target: "/")
    .linkStyle(.button)
    .linkDecoration(.light, hover: .heavy)
```

Or do I have that wrong 😅.

vdhamer:
I wrapped an Ignite Image in a Link.
Will the PR do anything here for non-text?

Sent from my iPhone

On 12 Nov 2024, at 18:12, Matt Hollyhead ***@***.***> wrote:

﻿

Works for me 🙂, thanks for the input @JPToroDev<https://github.com/JPToroDev>.

Being able to specify what you want in a single function like linkDecoration(.light, hover: .heavy) seems much more natural.

If folks then apply a linkStyle() for buttons as well, it feels like user error 😅.

@twostraws<https://github.com/twostraws>, Just so I have this right, do you mean if someone were to apply a linkDecoration, on top of a linkStyle it would be user error? like below?

Link("Click Me", target: "/")
    .linkStyle(.button)
    .linkDecoration(.light, hover: .heavy)

Or do I have that wrong 😅.

B4DG3R:
Hey @vdhamer 🙂, it shouldn't (as far as I am aware) at this point because the implementation uses bootstrap to modify the underline and hover properties on the link.  Unless there is something I am missing.

twostraws:
> Just so I have this right, do you mean if someone were to apply a linkDecoration, on top of a linkStyle it would be user error? like below?

Yes. It can do nothing for now, or we could return later to flag up a build warning.

vdhamer:
> Hey @vdhamer 🙂, it shouldn't (as far as I am aware) at this point because the implementation uses bootstrap to modify the underline and hover properties on the link. Unless there is something I am missing.

Ignore my remark: I just checked and found that I made an Image _behave_ like a link using an `.onClick` modifier instead of wrapping an Image inside a Link. That is not allowed (unlike `HTML <a></a>`).

B4DG3R:
@twostraws @JPToroDev I have made the updates to the link decoration work but I needed to merge the project main into my forkd main as there was a merge conflict.

After doing so and clearing the error I realised that there seems to be an issue with the @Page property and the project never stops running when you run it, hence no build folder is created.

I created a new test project to check it and the issues are present in a fresh project.  I think (although not 100%) that some of the issues may have been caused by PR #142 being merged in.

I have created a GitHub issue here detailing the issues I have found:

https://github.com/twostraws/Ignite/issues/144

I hope this helps and someone can take a look at it please.  I have taken a quick look but can't seem to work out how fresh projects are created, and the build issue has gone right over my head, although I am curious, time is limited so it may be better suited to the person who implemented the change 😅.

B4DG3R:
Apologies @twostraws @JPToroDev , I have just seen your Twitter post Paul 🤦‍♂️ 🫠

twostraws:
Sorry, things are in flux right now! There are a few more @JPToroDev pull requests to merge in, but hopefully we can get that done quickly and get everything back to being stable again. Apologies for the delay. In the meantime, are you happy for me to merge in your PR? It looks good to me!

B4DG3R:
It's cool no need to apologise 🙂, my initial worry was that it had been merged and no one was aware of the build errors.  I was going to raise some new evidence for the PR, but if you're happy with the updated code then I am happy for you to merge it 👍 🙌

Sources/Ignite/Elements/Link.swift
```diff
@@ -164,6 +164,50 @@ public struct Link: BlockElement, InlineElement, NavigationItem, DropdownElement
...
+            case .bold:
+                75
+            case .heavy:
+                100
+            }
+        }
+    }
+
+    /// The standard set of control attributes for HTML elements.
+    public var attributes = CoreAttributes()
```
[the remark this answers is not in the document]
JPToroDev: Yes, that should be no longer needed.

JPToroDev:
@twostraws @B4DG3R I think I know which build error you're talking about—a failure when building via Xcode, right? I fixed it, but it slipped through the cracks in my first PR. It's included in my latest.

B4DG3R:
Nice one @JPToroDev 🙌.  I have removed the `attributes` property @twostraws 👍.  Thanks both 🙂.

Sources/Ignite/Elements/Link.swift
```diff
@@ -284,6 +347,28 @@ public struct Link: BlockElement, InlineElement, NavigationItem, DropdownElement
...
+    /// - Parameters:
+    ///   - base: The `UnderlineProminence` for the base link style.
+    ///   - hover: The `UnderlineProminence` for the hover style.
+    /// - Returns: A new `Link` instance with the updated underline decoration.
+    func linkDecoration(_ base: UnderlineProminence = .heavy, hover: UnderlineProminence = .heavy) -> Self {
+        var copy = self
+        copy.baseDecoration = base
+        copy.hoverDecoration = hover
+        
+        // If there isn't already a role for this link,
```
JPToroDev, line 363: Is there a benefit to setting the role to `primary` here instead of leaving it as `default`?

Sources/Ignite/Elements/Link.swift
```diff
@@ -284,6 +347,28 @@ public struct Link: BlockElement, InlineElement, NavigationItem, DropdownElement
...
+        // add one automatically so it has sensible
+        // default button styling.
+        if copy.role == .default {
+            copy.role = .primary
```
JPToroDev, line 367: Is there a benefit to setting the role to `primary` here instead of leaving it as `default`?

B4DG3R:
@twostraws I have fixed the merge conflict on this, but I'm unsure if anything else needs changing due to the recent work that has been done.  I don't think so 🤔.  But met me know if there is 👍.

Sources/Ignite/Elements/Link.swift
twostraws: @B4DG3R Could you check this out, please? I'm keen to merge this one ASAP.

twostraws:
Now that the other PRs are merged, this one is next in line – could you give it a quick check with the latest changes, and let me know if we're all clear to merge? Thank you for your patience!

B4DG3R:
No worries, I should be able to sort that at some point today, and don't worry about the time taken 👍🙂.

B4DG3R:
Hey Paul, I am struggling a little with this one.  I have merged `Main` into my branch and the `Link` struct looks fine, but I am currently unable to get my site running with the new structure.  Hence I am unable to actually test it.

I finally managed to get it to a point where the project would build, although I still have a good chunk of it commented out, and when I run `ignite run --preview`, I get the error:

❌ This site specifies a custom subfolder, so it can't be previewed locally.

I did try and create a new test project to test it with, but it still seems to be creating a site with the old structure?

twostraws:
Push up any remaining code, and I'll merge and resolve it here. Thank you!

vdhamer:
> when I run `ignite run --preview`, I get the error:
> ❌ This site specifies a custom subfolder, so it can't be previewed locally.

The error message is related to the following setting defined in Site.url:
```
var url: URL = URL("https://www.vdhamer.com")
```
When previewing with `localhost` no subfolder is allowed. With a subfolder, you get that error message.
When generating for my actual website in a subfolder, I switch to using
```
var url: URL = URL("https://www.vdhamer.com/mydirectory")
```
for the build that gets uploaded/ftp'd to the server. Not great (because I sometimes forget to change the setting), but it works.

Yes, I read that twostraws will handle this. So this is mainly FYI.

B4DG3R:
It's all up to date, thanks for that, sorry I couldn't be of more help 👍

twostraws:
You've been very helpful, and very patient too – it's rare to have to sit alongside two very large other PRs. Thank you!

PR #142, opened 2024-11-12
SwiftUI-style `HTML` Protocol & Site Environment

JPToroDev:
# SwiftUI-style `HTML` Protocol & Site Environment

## Overview
This PR modernizes Ignite by introducing a SwiftUI-inspired `HTML` protocol and site environment system. These changes bring a more declarative, consistent, and ergonomic API while maintaining Ignite's core functionality.

### 1. New Protocol & Environment System
- Introduced new `HTML` protocol for SwiftUI-like view composition
- Added site environment system for dependency injection
- Implemented property wrappers (`@Environment`, `@Page`, `@Content`) for state management

### 2. Type Renaming Examples
Some key type names have been adjusted for clarity and to avoid conflicts:

| Old Name | New Name
|----------|----------|
| `HTML` | `HTMLDocument`
| `PageElement` | `HTML`
| `Body` | `HTMLBody`
| `Page` | `HTMLPage`

*Note: This is not an exhaustive list of renamed types.*

### 3. Simplified Type System
Some existing types became redundant with the new architecture. Examples include:
- `Component`
- `ContentPreview`

### Previous API
```swift
struct Home: StaticPage {
    var title = "Home"
    func body(context: PublishingContext) -> [BlockElement] {
        Text("Hello world!")
            .font(.title1)
    }
}

struct MyTheme: Theme {
    func render(page: Page, context: PublishingContext) -> HTML {
        HTML {
            Head(for: page, in: context)
            Body {
                page.body
                IgniteFooter()
            }
        }
    }
}
```

### New API
```swift
struct Home: StaticPage {
    var title = "Home"
    var body: some HTML {
        Text("Hello world!")
            .font(.title1)
    }
}

struct MyTheme: Theme {
    @Environment(\.siteConfiguration) private var config
    @Page private var page
    
    var body: some HTML {
        HTMLDocument {
            HTMLHead(for: page, with: config)
            HTMLBody {
                Group(page.body)
                IgniteFooter()
            }
        }
    }
}
```

## Benefits
1. **More Declarative**: The new API follows SwiftUI patterns, making it more intuitive for iOS developers
2. **Type Safety**: Improved type system helps catch errors at compile-time
3. **Dependency Injection**: Environment system provides cleaner dependency management
4. **Reduced Boilerplate**: Property wrappers eliminate repetitive context passing
5. **Better Naming**: More consistent and clear type names reduce confusion

## Migration Impact
- Existing pages will need to update their return types and context handling
- Theme implementations require property wrapper adoption
- Custom components may need to conform to new `HTML` protocol

Questions or feedback? Please let me know!


twostraws:
This is a really great pull request – thank you! Great code, great documentation, and tests too 🙌  I've read it through and it's very impressive. I have a handful of general questions that I'll add to specific lines shortly, but for other folks watching:

1. We're going to hold off updating Ignite Samples and other repositories until we get this sorted out. In the meantime I've pushed a tag that will give you some sort of stability.
2. Once this and some other big changes are merged in, we'll do a big sweep for SwiftLint etc all at once. This does mean it will be a little messy in the interim, but hopefully we'll get through it quickly!

Sources/Ignite/Elements/Link.swift
```diff
@@ -296,38 +273,58 @@ public struct Link: InlineElement, NavigationItem, DropdownElement {
...
+
+    /// Whether this link contains sensitive content that should be protected
+    private var isPrivacySensitive: Bool {
+        attributes.customAttributes.contains { $0.name == "privacy-sensitive" }
+    }
+
+    /// Renders a link with privacy protection enabled, encoding the URL and optionally the display content.
+    /// - Parameter context: The current publishing context.
+    /// - Returns: An HTML anchor tag with encoded attributes and content.
+    private func renderPrivacyProtectedLink(context: PublishingContext) -> String {
```
twostraws, line 305: I love this! But where is it actually exposed in the API? ("That will come in the next PR" is a perfectly fine response for this and any other comments; I know you've had to break this up and that's not easy.)

JPToroDev: My bad, that code shouldn't have been included here. It's part of an upcoming PR.

Sources/Ignite/Extensions/Array-DefinitelyNotAHack.swift
```diff
@@ -12,19 +12,20 @@ import Foundation
...
     }
-
-    public var attributes: CoreAttributes {
-        get { CoreAttributes() }
-        set { }
-    }
-
+    
     public func render(context: PublishingContext) -> String {
-        self.map { $0.render(context: context) }.joined()
```
twostraws, line 27: I'm not sure what's changing here.

Sources/Ignite/Elements/Modal.swift
```diff
@@ -119,36 +118,29 @@ public struct Modal: PageElement {
...
+    
     /// Renders this element using publishing context passed in.
     /// - Parameter context: The current publishing context.
     /// - Returns: The HTML for this element.
     public func render(context: PublishingContext) -> String {
         Group {
             Group {
                 Group {
-
-                    if header.isEmpty == false {
```
twostraws, line 131: This is a stylistic quirk of mine that I appreciate isn't shared by everyone – I find `== false` easier to parse.

JPToroDev: No problem with that.

Sources/Ignite/Elements/Image.swift
```diff
@@ -61,10 +61,10 @@ public struct Image: BlockElement, InlineElement, LazyLoadable {
     /// Allows this image to be scaled up or down from its natural size in
     /// order to fit into its container.
     /// - Returns: A new `Image` instance configured to be flexibly sized.
-    public func resizable() -> Self {
+    public func resizable() -> some BlockElement {
         var copy = self
         copy.attributes.classes.append("img-fluid")
-        return copy
+        return Group { copy }  // Return just the wrapped image with class
```
twostraws, line 67: Why wrap the image in a group? This returned `Self` before, allowing users to call `accessibilityLabel()` etc on it as well.

Sources/Ignite/Elements/ContentPreview.swift
```diff
@@ -1,83 +0,0 @@
...
-
-/// A protocol for customizing the layout of ContentPreview.
-public protocol ContentPreviewStyle {
-    func body(content: Content, context: PublishingContext) -> BlockElement
-}
-
-/// A wrapper around Card, specifically aimed at presenting details about
-/// some content on your site. This automatically links to your content page
-/// and adds in tags.
-public struct ContentPreview: BlockElement {
```
twostraws, line 18: This was a helpful element type to have around, for when folks wanted to show articles easily. Was it replaced by something else?

Sources/Ignite/Rendering/ResultBuilders/HTMLBuilder.swift
```diff
@@ -0,0 +1,308 @@
...
+    ///   - c0: First HTML element
+    ///   - c1: Second HTML element
+    ///   - c2: Third HTML element
+    ///   - c3: Fourth HTML element
+    ///   - c4: Fifth HTML element
+    ///   - c5: Sixth HTML element
+    ///   - c6: Seventh HTML element
+    ///   - c7: Eighth HTML element
+    /// - Returns: Combined array of HTML elements
+    public static func buildBlock<C0: HTML, C1: HTML, C2: HTML, C3: HTML, C4: HTML, C5: HTML, C6: HTML, C7: HTML>(
```
twostraws, line 270: This kind of code makes me think we have an equivalent to the old 10-view limit to SwiftUI, albeit 8 here. What are the implications of this? Would it be resolved through buiidPartialBlock() or similar?

Sources/Ignite/Elements/Image.swift
JPToroDev: I believe that has to do with an upcoming modifier called `imageFit()`. I can confirm if it's still necessary when I submit that PR—it was a modifier I created earlier in this journey.

Sources/Ignite/Framework/ElementTypes/HTML.swift
```diff
@@ -0,0 +1,288 @@
...
+        var attributes = attributes
+        attributes.styles.formUnion(values)
+        AttributeStore.default.merge(attributes, intoHTML: id)
+        return self
+    }
+    
+    /// Sets the `HTML` id attribute of the element.
+    /// - Parameter string: The ID value to set
+    /// - Returns: The modified `HTML` element
+    func htmlID(_ string: String) -> Self {
```
twostraws, line 207: Could you explain `id()` vs `htmlID()`? I can certainly see that `htmlID()` would produce `<p id="meh">woot</p>`, but what role does `id()` play?

Sources/Ignite/Elements/ContentPreview.swift
JPToroDev: Folks can now easily create their own reusable content preview views—just like you would in SwiftUI. To customize `ContentPreview`, you'd need to use `ContentPreviewStyle`, and that seems like a circuitous path to a result that's now part of the API and how you build elements.

Sources/Ignite/Framework/PropertyWrappers/Content.swift
```diff
@@ -0,0 +1,36 @@
...
+///     }
+/// }
+/// ```
+/// > Important: This property wrapper is only valid during content page rendering.
+/// The content context is managed automatically by the publishing system.
+@MainActor
+@propertyWrapper
+public struct Content {
+    public var wrappedValue: MarkdownContent {
+        ContentContext.current
```
twostraws, line 32: This might be another one of those not-quite-there-in-this-PR questions, but: is this strictly designed for use only in `ContentPage` structs? If so, given how it appears to be a singleton lookup, I wonder whether we could just expose this through a `content` property of a `ContentPage` protocol extension – not the SwiftUI way of doing things, but every `ContentPage` would need access to its content, so we're just saving a little time.

Sources/Ignite/Framework/PropertyWrappers/Tag.swift
```diff
@@ -0,0 +1,38 @@
...
+///             // Show articles matching the tag
+///         }
+///     }
+/// }
+/// ```
+/// > Important: This property wrapper is only valid in types that conform to `TagPage`.
+/// The tag context is managed automatically by the publishing system during rendering.
+@MainActor
+@propertyWrapper
+public struct Tag {
```
twostraws, line 29: My brain is stumbling a bit on this one – I can't quite tell whether `@Environment(\.tag) var currentTag` is a more natural spelling than `@Tag var currentTag`. Obviously the latter is much shorter!

Sources/Ignite/Framework/ElementTypes/HTML.swift
JPToroDev: Each view is assigned an `ID` behind the scenes, which is used to fetch its attributes. `id()` allows you to set an explicit `ID` for the view, because the automatic `ID` can change between builds. There aren't many cases where it's exposed—and I'll have to confirm if it's actually exposed at all now—but sometimes `ID` are used in the names of custom styles. Setting a custom `ID` would provide something stable for source control.

Sources/Ignite/Framework/Environment/EnvironmentStore.swift
```diff
@@ -0,0 +1,35 @@
...
+/// using the `update` method. For example:
+///
+/// ```swift
+/// let result = EnvironmentStore.update(newEnvironment) {
+///     // Code that needs the temporary environment
+///     return someValue
+/// }
+/// ```
+@MainActor
+final class EnvironmentStore {
```
twostraws, line 20: Does this allow SwiftUI-style environment nesting, where different parts of SwiftUI's view hierarchy have access to different environment values?

Sources/Ignite/Elements/Link.swift
twostraws: No problem; let's leave it in, and we can figure it out later.

Sources/Ignite/Framework/PropertyWrappers/Tag.swift
JPToroDev: `@Tag` is an interesting case. While it could technically be an environment key path since it's optional and therefore safe, I opted for a dedicated property wrapper. Why? It'd feel a bit misleading to have that key path show up as an available option in the wrong context. Having its own property wrapper scopes it better to where it actually belongs.

Sources/Ignite/Framework/Environment/EnvironmentStore.swift
JPToroDev: Yes, at the site environment level. View-based environment (i.e., having children inherit their parent's values when they lack one) is handled by CSS inheritance in upcoming PRs.

Sources/Ignite/Framework/PropertyWrappers/Content.swift
twostraws: I'm not sure the property wrapper would even be needed – maybe just this?

```swift
extension ContentPage {
    var content: MarkdownContent {
        ContentContext.current
    }
}
```

It makes the protocol a bit more magical because it gets `content` from ✨special places✨, but elsewhere in Swift we have things like `error` in catch blocks, `newValue` in `didSet`, etc, so automatic values aren't too surprising when they are effectively required.

Sources/Ignite/Elements/Image.swift
twostraws: Great! If it's not necessary any more then we can go back to `Self` being returned here, which gives a little more flexibility with the modifier order.

Sources/Ignite/Elements/ContentPreview.swift
twostraws: I'm thinking we should perhaps leave this one in for now – it makes it very easy to get a basic site up and running, with a home page, tag pages, etc. We can look at this again in the future once all the changes have landed. Is that okay?

Sources/Ignite/Framework/PropertyWrappers/Content.swift
JPToroDev: That's true. But those examples are present more in business logic APIs than in UI APIs—at least in the SwiftUI paradigm, which is what we're trying to pay homage to, right?

Sources/Ignite/Rendering/ResultBuilders/HTMLBuilder.swift
twostraws: This seems likely to be a stumbling block, so this should go on the "we need to research a better approach in the future" list. I think the preferred solution from Swift – and what SwiftUI uses – is pack iteration, but that's another rabbit hole. However, I don't think this gets in the way of merging this change as-is; we can figure it out post-merge.

Sources/Ignite/Elements/ContentPreview.swift
JPToroDev: Sure.

Sources/Ignite/Framework/PropertyWrappers/Tag.swift
twostraws: The other option is to follow the `ContentPage` suggestion and make it a property inside a protocol extension on `TagPage` – again, each tag page is pretty much guaranteed to want to know what tag it represents, so we're saving users some busywork.

Sources/Ignite/Elements/Image.swift
JPToroDev: Yes, let's do that. I encountered this issue with a new variant of `foregroundStyle()` recently, and I introduced a solution there that I think will also work here. 🙌🏼

Sources/Ignite/Framework/Environment/EnvironmentStore.swift
twostraws: Understood – thank you!

Sources/Ignite/Framework/PropertyWrappers/Content.swift
twostraws: Yes, but I'm okay with that – it's a pragmatic choice to make coding a little easier, and I don't think it really has much of a parallel in SwiftUI. If users find it weird going forward, removing it and asking them to make the content explicit wouldn't be too hard. If there were lots of times when folks were making content pages that didn't actually need to know what content to render, I'd think otherwise 😅

JPToroDev: OK, I see your point. Let's go with that for `ContentPage` and `TagPage` then!

Sources/Ignite/Framework/ElementTypes/HTML.swift
twostraws: Is the new `id` property for internal use only? If so, I'd be inclined to make it internal and hidden away as something like `uniqueIdentifier` and make `htmlID()` back into `id()` because it's more likely to be used.

Sources/Ignite/Rendering/ResultBuilders/HTMLBuilder.swift
JPToroDev: Agreed—sounds great to me!

Sources/Ignite/Framework/ElementTypes/HTML.swift
JPToroDev: Yes, for internal use only. I agree `htmlID()` is more likely to be used, but I think naming it `id()` and then having a modifier with a different name that works like SwiftUI's `id()` modifier could be confusing. But I can get behind this!

twostraws: SwiftUI has "interesting" ideas here, and I'm not too wedded to their approach – `id()` and `tag()` are a bit confused over there, with things like [resolveSymbol(id:)](https://developer.apple.com/documentation/swiftui/graphicscontext/resolvesymbol(id:)) actually relying on [View.tag(_:)](https://developer.apple.com/documentation/swiftui/group/tag(_:)), for example.

If what is currently set using `id()` is definitively for internal use only, then I think "having a modifier with a different name that works like SwiftUI's id() modifier could be confusing" is going to be okay – most folks won't see the modifier, and perhaps even *can't* see the modifier if it's marked as being internal.

From what I can see, the `id` setter is currently a no-op.

> but sometimes ID are used in the names of custom styles

This definitely feels more like HTML's id attribute than an internal identifier.

JPToroDev: All good points. Happy to go with your recommendation here!

twostraws: Great! We can always return to this again in the future as needed. I'll write a sum up comment now.

twostraws:
Thank you for all the clarifications! Based on what we've discussed:

- `renderPrivacyProtectedLink` – you're going to leave this as-is for now; it's not used currently, but will be soon.
- `try await render(site.homePage, isHomePage: true)` – leave the `await` out for now, but I want to revisit this once all other changes are merged in because I appreciate folks want to fetch dynamic data during a build.
- `self.map { $0.render(context: context) }.joined()` – we can put this back later, or you can add a commit to change it back now if you want.
- `if header.isEmpty == false {` – I think I might change this back later on 🙈
- `return Group { copy }` – please add a commit removing the group, and using `Self` for the return value. We can revisit this in the future if needed.
- Please reinstate `ContentPreview`; we might end up removing this once things have settled, but I'd rather keep it for the time being.
- `buildBlock(……………)` – we'll merge this in for now, and come back to it soon. I'm not keen to have an 8-child limit if possible.
- `htmlID()` – please put this back to `id()`; you're welcome to choose some other internal name for the other unique identifier used for attributes etc. `internalIdentifier`?
- `@Tag` – Please add a commit to push this and `@Content` over to properties provided through protocol extensions for their respective protocols.

This is such a great set of changes, and I have no doubt others are keen to try it all out 🙌

JPToroDev:
Sounds great, Paul. I'll have this all squared away tonight!

JPToroDev:
@twostraws All set. 🫡

twostraws:
Perfect – thank you!

PR #143, opened 2024-11-13
# `Animatable`, `Animation`, and the New `animation()` Modifier

JPToroDev:
# `Animatable`, `Animation`, and the New `animation()` Modifier

## Overview
This PR introduces a powerful animation system to Ignite, enabling developers to create sophisticated CSS animations using type-safe Swift syntax. The system provides a comprehensive suite of features including animation triggers, keyframe generation, and pre-built animations.

### 1. Type-Safe Animation API
The new API provides a fluent, Swift-native interface for defining animations:

```swift
Text("Hello, world!")
     .animation(.fadeIn().duration(3), on: .appear)
     .animation(.color(from: "red", to: "blue"), on: .click)
     .animation(.scale(from: 1, to: 1.5).duration(1), on: .click)
```

### 2. Animation Triggers
The system supports multiple ways to initiate animations:

| Trigger | Description |
|---------|-------------|
| `.appear` | Activates when element enters viewport |
| `.hover` | Activates on mouse hover |
| `.click` | Activates on element click/tap |

## Key Features
1. **Type-Safe API**: 
  - Compile-time checking for animation properties
  - Auto-completion support
  - Clear parameter requirements

2. **Multiple Animation Triggers**:
  - Viewport-based animations
  - User interaction triggers
  - Flexible trigger combinations

3. **Rich Animation Library**:
  - Pre-built animation effects
  - Customizable duration and timing
  - Support for keyframe animations

## Benefits
1. **Developer Experience**: Intuitive Swift syntax for CSS animations
2. **Type Safety**: Catch animation configuration errors at compile-time
3. **Flexibility**: Multiple trigger options and animation combinations
4. **Maintainability**: Clear, declarative animation definitions

## Migration Impact
- Compatible with existing Ignite components
- No breaking changes to current animation implementations
- Gradual adoption possible

## Next Steps
- [ ] Add comprehensive documentation and tests
- [ ] Create animation cookbook with examples
- [ ] Add more pre-built animations
- [ ] Implement advanced keyframe support

Questions or feedback? Please let me know!


twostraws:
I'm going through the code now – can you confirm your version definitely builds? I'm getting quite a few build errors here.

Sources/Ignite/Elements/HTMLHead.swift
```diff
@@ -91,6 +91,7 @@ public struct HTMLHead: HTMLRootElement {
             analyticsSnippet
         }
     
+        MetaLink.animationCSS
```
twostraws, line 94: If we always need the animation CSS file – which I'm not sure is true, if a site does not use animations – then we should probably just include the file directly in the main CSS, to transfer it as one.

Sources/Ignite/Extensions/URL-SelectDirectories.swift
```diff
@@ -9,37 +9,28 @@ import Foundation
...
-    /// When building a package, the website is built at the source URL (and both URLs are equal).
-    /// When building a MacOS app, the website is built in a subdirectory of the app's sandbox.
     /// - Parameter file: path of a Swift source file to find source root directory by scanning path upwards.
     /// - Returns tupple containing source URL and URL where output is built.
     public static func selectDirectories(from file: StaticString) throws -> SourceBuildDirectories {
         var currentURL = URL(filePath: file.description)
-
-        repeat {
+        
+        while currentURL.path != "/" {
```
twostraws, line 17: This is a part of the code @vdhamer has worked on a lot – Peter, can you take a look at this, please? I want to make sure whatever changes you want to introduce still work here, or at least can be applied over this PR once merged.

Sources/Ignite/Framework/Animations/AnimationBuilder.swift
```diff
@@ -0,0 +1,95 @@
...
+///
+/// `AnimationBuilder` provides the foundation for Ignite's animation DSL, allowing animations
+/// to be composed from simple building blocks into more complex sequences. It handles the conversion
+/// of various animation types into `ResolvedAnimation` instances that can be rendered to CSS.
+@resultBuilder
+public struct AnimationBuilder {
+    /// Builds a single animation from one or more animation components.
+    /// Converts various animation types into a resolved animation for rendering.
+    public static func buildBlock(_ components: any Animation...) -> some Animation {
+        guard let component = components.first else { return ResolvedAnimation() }
```
twostraws, line 20: This method returns several different possibilities; it'd be helpful if you could add a comment before each condition explaining what is being sent back and why 👍

Sources/Ignite/Framework/Animations/AnimationClassGenerator.swift
```diff
@@ -0,0 +1,216 @@
...
+        """
+    }
+    
+    /// Generates the base CSS class containing common properties
+    /// - Returns: A string containing the base CSS class definition
+    func buildBaseClass() -> String {
+        var properties = Set<String>()
+        
+        if triggerMap[.click] != nil || triggerMap[.hover] != nil {
+            properties.insert("cursor: pointer")
```
twostraws, line 68: A great touch! 👌

Sources/Ignite/Framework/Animations/AnimationClassGenerator.swift
```diff
@@ -0,0 +1,216 @@
...
+        """
+    }
+    
+    /// Generates all trigger-specific CSS classes
+    /// - Returns: A string containing all trigger-specific CSS class definitions
+    private func buildTriggerClasses() -> String {
+        triggerMap.compactMap { trigger, animation in
+            switch trigger {
+            case .click:
+                return buildClickClass(animation)
```
twostraws, line 192: Nit: For code consistency, I'd suggest dropping the `return` keywords in this `switch`.

Sources/Ignite/Framework/Animations/AnimationCurve.swift
```diff
@@ -0,0 +1,37 @@
...
+    /// Starts slow, speeds up in the middle, then slows down
+    case easeInOut
+    /// Spring-like animation with customizable damping and velocity
+    case spring(dampingRatio: Double, velocity: Double)
+    
+    case custom(String)
+    
+    var cssValue: String {
+        switch self {
+        case .linear: return "linear"
```
twostraws, line 27: Nit: Same here; drop the `return`s.

Sources/Ignite/Framework/Animations/AnimationManager.swift
```diff
@@ -0,0 +1,87 @@
...
+    /// Registers any animation type for a specific element.
+    ///
+    /// This method attempts to resolve the animation into a `ResolvedAnimation` through
+    /// various means before registration.
+    ///
+    /// - Parameters:
+    ///   - animation: The animation to register
+    ///   - elementID: The unique identifier of the element to animate
+    func registerAnimation(_ animation: some Animation, for elementID: String) {
+        if let resolved = animation as? ResolvedAnimation {
```
twostraws, line 47: Again, some comments here explaining what each `if` checked for and why it returns would be helpful, to guide reading flow. Thank you!

Sources/Ignite/Framework/Site.swift
```diff
@@ -104,61 +104,61 @@ public protocol Site {
     func publish(from file: StaticString, buildDirectoryPath: String) async throws
 }
 
-extension Site {
+public extension Site {
```
twostraws, line 107: My favorite kind of change 🙂

Sources/Ignite/Framework/Animations/AnimationCurve.swift
```diff
@@ -0,0 +1,37 @@
...
+// AnimationValue.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+import Foundation
+
+/// Defines the available timing functions for animations
+public enum AnimationCurve {
```
twostraws, line 11: Is it possible we could bring this closer to the SwiftUI style of animation curves? For example, my favorite custom SwiftUI animation curve is ease-in-out-back, which mimics cartoon-style animation – the object starts by moving in the wrong direction, then overshoots its target with a spring, the bounces back. In SwiftUI code it'd be this:

```swift
extension Animation {
    static var easeInOutBack: Animation {
        Animation.timingCurve(0.5, -0.5, 0.5, 1.5)
    }
}
```

If we followed that approach here, it would mean having an extra case where we can manually specify each of the control points of a cubic Bézier curve.

Sources/Ignite/Framework/Animations/StandardAnimation.swift
```diff
@@ -0,0 +1,56 @@
...
+/// transitions of individual CSS properties like opacity, transform, or color.
+public struct StandardAnimation: Animation, Animatable {
+    /// The starting value for the animated property
+    var from: String
+    
+    /// The ending value for the animated property
+    var to: String
+    
+    /// The CSS property to animate (e.g., "opacity", "transform", "background-color")
+    var property: String
```
twostraws, line 22: Is a `String` the right choice here, as opposed to an enum of animatable properties?

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+        }
+        
+        return StandardAnimation(property: "transform", from: "\(property)(0)", to: "\(property)(\(to))")
+            .duration(0.35)
+            .timing(.easeOut)
+    }
+    
+    /// Creates a rotation animation by specified degrees.
+    static func rotate(degrees: Double) -> StandardAnimation {
+        StandardAnimation(property: "transform", from: "rotate(0deg)", to: "rotate(\(degrees)deg)")
```
twostraws, line 174: Seeing all these strings everywhere makes me a little uneasy. Is it the case that users will never see the stringly-typed API and will instead rely on our type-safe wrappers?

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+public extension Animation where Self == StandardAnimation {
+    /// Creates a fade-in animation that transitions opacity from 0 to 1.
+    static func fadeIn() -> StandardAnimation {
+        StandardAnimation(property: "opacity", from: "0", to: "1")
+            .duration(0.35)
+            .timing(.easeOut)
+    }
+    
+    /// Creates a fade-out animation that transitions opacity from 1 to 0.
+    static func fadeOut() -> StandardAnimation {
```
twostraws, line 136: Someone is bound to ask: how do I make a thing flip _and_ fade in over 1 second?

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+            AnimationManager.shared.registerAnimation(resolvedAnimation, for: elementID)
+            return resolvedAnimation.name
+        }
+        return ""
+    }
+}
+
+public extension Animation where Self == StandardAnimation {
+    /// Creates a fade-in animation that transitions opacity from 0 to 1.
+    static func fadeIn() -> StandardAnimation {
```
twostraws, line 129: I'm wondering whether we should follow SwiftUI here and make these methods into properties if they have no parameters, e.g. `.transition(.opacity)` and `.transition(.slide)`, but `.transition(.move(edge: .top))`.

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+        StandardAnimation(property: "transform", from: "rotate(0deg)", to: "rotate(\(degrees)deg)")
+            .duration(0.35)
+            .timing(.easeOut)
+    }
+    
+    /// Creates a scale animation between two values.
+    static func scale(from: Double = 0.8, to: Double = 1.0) -> StandardAnimation {
+        StandardAnimation(property: "transform", from: "scale(\(from))", to: "scale(\(to))")
+            .duration(0.35)
+            .timing(.spring(dampingRatio: 0.5, velocity: 0.1))
```
twostraws, line 183: Providing default durations and timing curve is fine, but my instinct is to use the same default duration and timing curve for each built-in animation unless it's something unusual, e.g. `bounce()`. What do you think?

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+        StandardAnimation(property: "filter", from: "blur(0px)", to: "blur(\(radius)px)")
+            .duration(0.35)
+            .timing(.easeInOut)
+    }
+    
+    /// Creates a bouncing animation.
+    static func bounce() -> StandardAnimation {
+        StandardAnimation(property: "transform", from: "translateY(0)", to: "translateY(-20px)")
+            .duration(0.5)
+            .timing(.custom("cubic-bezier(0.36, 0, 0.66, -0.56)"))
```
twostraws, line 211: We should retain `.custom("…")` for times when someone wants something very specific, e.g. the "steps" animation curve, but for just general cubic Bézier curves I think we should follow SwiftUI in just accepting four doubles. (Suggested elsewhere, but an example is here.)

Sources/Ignite/Framework/Animations/AnimationCurve.swift
```diff
@@ -0,0 +1,37 @@
...
+    
+    case custom(String)
+    
+    var cssValue: String {
+        switch self {
+        case .linear: return "linear"
+        case .easeIn: return "ease-in"
+        case .easeOut: return "ease-out"
+        case .easeInOut: return "ease-in-out"
+        case .spring(let damping, let velocity):
```
twostraws, line 31: I like how SwiftUI's animations provide some built-in springs automatically, for folks who either don't understand or don't _want_ to understand damping and velocity. Could we add a couple of built-in springs too – `.snappy` and `.bouncy`, etc?

Here's what SwiftUI uses:

```swift
static var smooth: Spring {
    get { smooth() }
}

static func smooth(duration: TimeInterval = 0.5, extraBounce: Double = 0.0) -> Spring {
    Spring(duration: duration, bounce: extraBounce)
}

static var snappy: Spring {
    get { snappy() }
}

static func snappy(duration: TimeInterval = 0.5, extraBounce: SDouble = 0.0) -> Spring {
    Spring(duration: duration, bounce: 0.15 + extraBounce)
}

static var bouncy: Spring {
    get { bouncy() }
}

static func bouncy(duration: TimeInterval = 0.5, extraBounce: Double = 0.0) -> Spring {
    Spring(duration: duration, bounce: 0.3 + extraBounce)
}
```

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+        default: ("translateX", "-100%")
+        }
+        
+        return StandardAnimation(property: "transform", from: "\(property)(0)", to: "\(property)(\(to))")
+            .duration(0.35)
+            .timing(.easeOut)
+    }
+    
+    /// Creates a rotation animation by specified degrees.
+    static func rotate(degrees: Double) -> StandardAnimation {
```
twostraws, line 173: Are folks going to expect this to have some kind of anchor, similar to SwiftUI's rotations? I _believe_ the CSS equivalent is `transform-origin`. (This might be too much to tackle in this PR.)

twostraws:
Thank you for another fantastic pull request! I've read through the code several times, and added comments with some small notes – some general spelling questions, a few requests for comments here and there, etc. The only things I really have outstanding questions on are:

1. Is it possible to replace some of the stringly typed APIs with enums? I don't even know if it's necessary because this API might not be exposed to developers, but if they need to write things like "transform", "background-color", etc, then I think an enum would be a good idea.
2. I think some places might work better moved towards SwiftUI a little more, namely built-in spring types, using properties rather than parameter-less methods, and letting users create Bézier curves through four doubles rather than writing out the whole string. If adding anchors for things is also possible trivially, that would also be welcome.

Sources/Ignite/Elements/HTMLHead.swift
JPToroDev: The file wouldn't be needed if a site doesn't use animations. We could check to see if the `animations` dictionary inside `AnimationManager` is empty and if so, forego generating it.

Sources/Ignite/Extensions/URL-SelectDirectories.swift
JPToroDev: My changes brought out what seemed to be a preexisting issue: when "Package.swift" wasn't found, the loop wouldn't stop, and it'd cause a crash. And even after fixing that, the code inside the following conditional was never called:

```swift
let buildDirectory: String = NSHomeDirectory() // app's home directory for a sandboxed MacOS app
if buildDirectory.contains("/Library/Containers/") {
    let buildDirectoryURL: URL = URL(filePath: buildDirectory)
    return SourceBuildDirectories(source: buildDirectoryURL, build: buildDirectoryURL)
}     
```

Using `#filepath` instead of `#file` allowed for the original directory to be accessed regardless of whether we were building in Xcode or the command line, so the above could be safely removed.

Sources/Ignite/Framework/Animations/AnimationClassGenerator.swift
JPToroDev: Sure, no problem.

Sources/Ignite/Framework/Animations/AnimationManager.swift
JPToroDev: Sure thing.

Sources/Ignite/Framework/Site.swift
JPToroDev: Yeah, those easy wins are hard to be beat 🤣

Sources/Ignite/Modifiers/Animation-Modifier.swift
```diff
@@ -0,0 +1,90 @@
...
+
+import Foundation
+
+public extension HTML {
+    /// Applies an animation to an HTML element.
+    /// - Parameters:
+    ///   - animation: The animation to apply
+    ///   - trigger: When the animation should occur (click, hover, or appear)
+    /// - Returns: The modified HTML element with animation support
+    func animation(_ animation: Animation, on trigger: AnimationTrigger) -> Self {
```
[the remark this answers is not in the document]
JPToroDev: Yes, the same animation with two different triggers would require two modifiers at the moment. Good idea! I don't see why making `trigger` an option set would be a problem—I'll give it a shot.

Sources/Ignite/Framework/Animations/AnimationCurve.swift
JPToroDev: I like this! What other curves would you want to see?

Sources/Ignite/Framework/Animations/StandardAnimation.swift
JPToroDev: You're right. I have an enum called `Property` that's exhaustive, and I'm currently using it in an alternate init for `AttributeValue`. I don't think I included it in my last PR. I'll add it with this one and use it here.

Sources/Ignite/Extensions/URL-SelectDirectories.swift
twostraws: @vdhamer was particularly interested in using Ignite inside an existing macOS project as a package, e.g. a macOS SwiftUI app that generates a website. I think this is likely the trigger for that condition?

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: The `Property` enum can clear up half of it. Do you see that being an issue for CSS values? Am I following you correctly here?

Sources/Ignite/Framework/Animations/AnimationCurve.swift
twostraws: I mentioned a few in a later comment – `bouncy`, etc – but at the very least need some kind of custom case that accepts four Bézier curve control points. Perhaps `.timingCurve(_:_:_:_:)`?

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: To change the duration you could do `fadeOut().duration(3)`. For compound animations, you can use the `KeyframeAnimation` type and get crazy. That's the main reason I wanted to rework the animation system 😁

The API works like this—which reminds me I need to create an init for `StandardAnimation` that takes a single property.

```swift
struct MyAnimation: Animation {
    var body: some Animation {
        KeyframeAnimation("crazy-colors") {
            Frame("0%") {
                StandardAnimation(property: "background-color", from: "blue", to: "yellow")
            }
            Frame("50%") {
                StandardAnimation(property: "background-color", from: "yellow", to: "red")
            }
            Frame("100%") {
                StandardAnimation(property: "background-color", from: "red", to: "purple")
            }
        }
        .duration(1.0)
        .timing(.spring(dampingRatio: 0.5, velocity: 0.1))
    }
}
```

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Good idea, I like that better.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Totally agree!

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Makes sense. Will work this out 🙌🏼

Sources/Ignite/Framework/Animations/AnimationCurve.swift
JPToroDev: Sure!

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Let me investigate this. I do think it might not be a bad idea to tackle some of these things down the line.

Sources/Ignite/Framework/Animations/Animation.swift
twostraws: I'm afraid I don't know enough about CSS animations here, but are those frame positions only ever specified as percentages? If so, an integer rather than a string might work better?

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: So instead of separating properties and values, would the ideal system be something like this: `.transform(.rotate(0))`, `.color(.red)`, `.font("San Francisco")`?

twostraws: Possibly; I'd need to think some more, because Apple takes a few different approaches here. I think we should park this and return in a future PR.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: "from" and "to" are also valid, but they map to 0% and 100%—so yes, let's do an `Int` instead!

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Sounds good. What are some of Apple's other approaches that you think could work, just so I can have them turning in the back of my head?

twostraws: With zero thought – these are just rattling around in my head! – there are several that spring to mind.

First, how Swift has all sorts of specialized types, e.g. mapping `AsyncSequence` produces `AsyncMapSequence`, filtering it produces `AsyncFilterSequence`, prefixing it produces `AsyncPrefixSequence`, etc. In the case of transformations, it might mean having a handful of nested specialized types for different animations, e.g.:

```swift
KeyframeAnimation("crazy-colors") {
    Frame(0) {
        BackgroundColorAnimation(from: .blue, to: .yellow)
    }
    Frame(0.5) {
        ScaleAnimation(from: 0, to: 1)
    }
    Frame(1) {
        RotationAnimation(from: .degrees(0), to: .degrees(180))
    }
}
.duration(3)
```

It's pretty wordy, though, which lead me onto the second idea: SF Symbols animations have various overloads that only work with specific animation types.  As an example, this code is allowed:

```swift
struct ContentView: View {
    @State private var isFavorite = false

    var body: some View {
        Button("Bounce Me", systemImage: "mail.stack") {
            isFavorite.toggle()
        }
        .symbolEffect(.bounce, value: isFavorite)
    }
}
```

However, if you look at the code completion options in place of `.bounce` and try `.appear`, you'll see it doesn't work – you get the error "Instance method 'symbolEffect(_:options:value:)' requires that 'AppearSymbolEffect' conform to 'DiscreteSymbolEffect'".  Similarly, if you try doing `.contentTransition(.symbolEffect(.bounce))` it won't work, whereas `.contentTransition(.symbolEffect(.replace))` will.

So, an option here might be to have an `Animation` protocol extension method called `animate()` that is overloaded: if the first parameter is a `DoubleAnimationProperty` (for example), then it would take two Doubles, whereas a `ColorAnimationProperty` would take two colors, etc. This would yield API like this:

```swift
func animate(_ property: Animation.DoubleProperty, from: Double, to: Double)
func animate(_ property: Animation.ColorProperty, from: Color, to: Color)
func animate(_ property: Animation.AngleProperty, from: Angle, to: Angle)
```

(We don't have an `Angle` type right now, but I suspect it's a good idea.)

And at the call site:

```swift
KeyframeAnimation("crazy-colors") {
    Frame(0) {
        animate(.backgroundColor, from: .blue, to: .yellow)
    }
    Frame(0.5) {
        animate(.scale, from: 0, to: 1)
    }
    Frame(1) {
        animate(.rotation, from: .degrees(0), to: .degrees(180))
    }
}
.duration(3)
```

A middle ground might be to have lots of `Animation` protocol extension methods, yielding code like this:

```swift
KeyframeAnimation("crazy-colors") {
    Frame(0) {
        backgroundColor(from: .blue, to: .yellow) // this one doesn't read well; `colorBackground()` at least reads like an action.
    }
    Frame(0.5) {
        scale(from: 0, to: 1)
    }
    Frame(1) {
        rotate(from: .degrees(0), to: .degrees(180))
    }
}
.duration(3)
```

If we were to go down the route of combined animations, then the third option is to go back towards SwiftUI style, like this:

```swift
KeyframeAnimation("crazy-colors")
    .keyframe(0)
        .backgroundColor(from: .blue, to: .yellow)
    .keyframe(0.5)
        .scale(from: 0, to: 1)
    .keyframe(1)
        .rotation(from: .degrees(0), to: .degrees(180))
```

That isn't quite right because the modifier chain all seems to apply to the keyframe animation rather than each part modifying the current frame – the extra indents might make it look different, but it's not. So maybe this:

```swift
KeyframeAnimation("crazy-colors") {
    Frame(0)
        .backgroundColor(from: .blue, to: .yellow)

    Frame(0.5)
        .scale(from: 0, to: 1)

    Frame(1)
        .rotate(from: .degrees(0), to: .degrees(180))
}
.duration(3)
```

That variant is relatively like on syntax, clearly attaches the animation type to the frame, and also leaves open the possibility of multiple animations by adding more modifiers to the `Frame`.

I think I'd need to get a better idea of all the various things we might want to animate, to see how much of a linguistic mess we get into with verbs!

Again, I haven't had time to really think these through, so apologies if they sound unhinged – it's 1:45am here 😅

twostraws:
Thank you for all the replies! Based on the discussion, you're going to:

- Add a commit to skip generating/including any animations code if there aren't any.
- Add a handful of extra comments here and there.
- Drop `return` in a handful of places.
- Investigate whether we can add animation triggers as an option set.
- Bring in your `Property` enum (presumably `Animation.Property`?), so we can remove some string-iness from the API.
- Change animation functions such as `fadeIn()` that have no parameters into computed properties.
- Switch to the same default timing and curves everywhere that's appropriate – bounce, shake, and flip should probably be excepted.
- Add some built-in spring types to mimic snappy and bouncy from SwiftUI, and also add an animation timing curve case that accepts four doubles for easier custom curves.
- Switch keyframe positions to numbers. SwiftUI doesn't have much (any?) prior art for percentages – a range of through through 1 as doubles is much more common. I'm not sure if it's worth adding overloads for an enum called `KeyframeAnimation.Position` that contains the cases `from` and `to` but internally just map to 0 and 1? (I suspect not.)


Let's punt for the future:

- Adding an anchor for various animations. If this happens in the future it will just be an extra parameter with a default value, so it shouldn't break anything.
- Whether CSS values in animations should have specific types such as `.color(.red)`.

We should tread carefully around the "/Library/Containers/" stuff, because this has been a problem point in the past. It would be helpful if @vdhamer could write some tests that validate this part of Ignite all behaves correctly for various inputs, so any future changes are validated correctly for both standalone sites and sites that are incorporated into apps somehow. If you're not sure, I'd suggest withdrawing this change from the PR here, and Peter can look into it further in his own PR.

Unless you have another spark of brilliance, I'd also suggest punting the keyframe animation changes too – it's worth exploring some other spellings there to see if we can find something that really feels natural.

I have one outstanding question, regarding this: "For compound animations, you can use the KeyframeAnimation type and get crazy. That's the main reason I wanted to rework the animation system." I was thinking folks wouldn't be asking about "flip THEN fade in," but instead "flip WHILE fading in" - an animation that does two or three things at once. (Definitely one for the "punt" list.)

Thank you!

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Not unhinged at all—in fact, I'm impressed you're that coherent at 1:45am!

> ```swift
> KeyframeAnimation("crazy-colors")
>     .keyframe(0)
>         .backgroundColor(from: .blue, to: .yellow)
>     .keyframe(0.5)
>         .scale(from: 0, to: 1)
>     .keyframe(1)
>         .rotation(from: .degrees(0), to: .degrees(180))
> ```
> That isn't quite right because the modifier chain all seems to apply to the keyframe animation rather than each part modifying the current frame – the extra indents might make it look different, but it's not...

I think the fully realized version might look something like this:

```swift
KeyframeAnimation("crazy-colors")
    .keyframe(0) {
        $0.backgroundColor(from: .blue, to: .yellow)
        $0.scale(from: 0, to: 1)
    }
    .keyframe(100) {
        $0.rotation(from: .degrees(0), to: .degrees(180))
    }
```

Which I quite like. What do you think?

JPToroDev:
@twostraws Sounds good. I'll start tackling this all in the morning!

I'll put back the "/Library/Containers/" check, but will leave the revised loop above it as is to avoid a build failure via Xcode. 

> I was thinking folks wouldn't be asking about "flip THEN fade in," but instead "flip WHILE fading in" - an animation that does two or three things at once.

That's exactly what I meant. We can do stuff like this:

```swift
struct MyAnimation: Animation {
    var body: some Animation {
        KeyframeAnimation("bounce") {
            Frame("0%") {
                StandardAnimation(property: "background-color", value: "blue")
                StandardAnimation(property: "color", value: "yellow")
            }
            Frame("100%") {
                StandardAnimation(property: "background-color", value: "red")
                StandardAnimation(property: "color", value: "white")
            }
        }
        .duration(10.0)
        .timing(.spring(dampingRatio: 0.5, velocity: 0.1))
    }
}
```

I ironed out a bug that was preventing this from working that will be included in my next commit. 🙌🏼

Sources/Ignite/Framework/Animations/Animation.swift
twostraws:
```swift
KeyframeAnimation("crazy-colors")
    .keyframe(0) {
        $0.backgroundColor(from: .blue, to: .yellow)
        $0.scale(from: 0, to: 1)
    }
    .keyframe(100) {
        $0.rotation(from: .degrees(0), to: .degrees(180))
    }
```

That might look better with shorthand syntax, but I'm wondering whether it would be more natural as builder syntax:

```swift
KeyframeAnimation("crazy-colors")
    .keyframe(0) { content in
        content
            .backgroundColor(from: .blue, to: .yellow)
            .scale(from: 0, to: 1)
    }
    .keyframe(100) {
        $0.rotation(from: .degrees(0), to: .degrees(180))
    }
```

So, where it's just one thing the syntax is what you proposed, but you can chain them like SwiftUI modifiers.

We'd also need to think about naming for the animation methods. I think I prefer verbs everywhere it's practical, but it's not required – `scale()` is fine, `rotation()` might be better as `rotate()`, but `backgroundColor()` should just stay as-is. For things that suggest render changes only, SwiftUI uses `…Effect()`, like `scaleEffect()` and `rotationEffect()`, but I think that doesn't quite fit here because these aren't simple rendering effects – the underlying CSS model is changing. Elsewhere, they use verbs like `disabled()`, adjectives like `searchable()`, and nouns like `padding()` based on whatever seems natural.

Again, this can happily be pushed back to a later discussion – I've already had four folks reach out to me to say they just started with Ignite and the docs aren't right, so I think we should get this and any other commits merged, and I can start work on getting things back in sync in this repository, in samples, and in the starter template.

twostraws:
Sounds good to me.

> That's exactly what I meant. We can do stuff like this:

Ah., great; thank you 👍 

Reading this again The Morning After has left me with a couple more questions/notes, which I'll add to specific lines now.

Sources/Ignite/Framework/Animations/Animation-Frame.swift
```diff
@@ -0,0 +1,53 @@
...
+/// `Frame` represents a specific point in an animation timeline, defining one or more property
+/// transformations at a given position (percentage or keyword).
+///
+/// Example:
+/// ```swift
+/// Frame("50%") {
+///     StandardAnimation(property: "transform", from: "scale(1)", to: "scale(1.2)")
+/// }
+/// ```
+public struct Frame: Animation, Animatable {
```
twostraws, line 21: I think this should be `Keyframe`. It's clearer, yes, but `Frame` is quite a common word for layout stuff, and it might cause some confusion elsewhere. I rather wish all these things could be nested inside something like `Animation`, giving `Animation.Frame`, but that would make it supremely unwieldy at the call site!

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+public extension Animation {
+    /// A unique identifier generated from the animation type name
+    static var id: String {
+        String(describing: self).truncatedHash
+    }
+    
+    /// A kebab-case name combining the type name and unique identifier
+    var name: String {
+        let typeName = String(describing: type(of: self))
+            .replacingOccurrences(of: "Style", with: "")
```
twostraws, line 49: Elsewhere we're using Swift's own `replacing()` rather than Foundation's `replacingOccurrences()`; I think we should stay consistent.

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+        let kebabCase = typeName.replacingOccurrences(
+            of: "([a-z0-9])([A-Z])",
+            with: "$1-$2",
+            options: .regularExpression
+        ).lowercased()
```
twostraws, line 54: For the regex approach here, we can use `replacing()` with a regex – see `Text.init(markdown:)` for an example.

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+        return "\(kebabCase)-\(UUID().uuidString.truncatedHash)"
+    }
+}
+
+public extension Animation {
+    /// Adjusts the animation speed by modifying its duration.
+    /// - Parameter speed: The speed multiplier. Values greater than 1 increase speed, less than 1 decrease it.
+    func speed(_ speed: Double) -> Self {
+        if var copy = self as? any Animatable {
+            duration(copy.duration / speed)
```
twostraws, line 65: Should this assign to `copy.duration` rather than calling the method? (There are a bunch of other build warnings elsewhere – "Result of call to 'class' is unused", "Result of call to 'class' is unused", etc.)

Sources/Ignite/Framework/Animations/StandardAnimation.swift
```diff
@@ -0,0 +1,56 @@
...
+// See LICENSE for license information.
+//
+
+import Foundation
+
+/// A basic animation type that transitions a single CSS property from one value to another.
+///
+/// `StandardAnimation` represents the simplest form of animation in Ignite, handling
+/// transitions of individual CSS properties like opacity, transform, or color.
+public struct StandardAnimation: Animation, Animatable {
```
twostraws, line 14: The name `StandardAnimation` makes me think others are nonstandard, rather than just different. How about `SimpleAnimation`? Or `ValueAnimation`, `TransitionAnimation`, etc?

Sources/Ignite/Modifiers/Animation-Modifier.swift
```diff
@@ -0,0 +1,78 @@
...
+        return copy.animation(animation, on: .appear)
+    }
+    
+    /// Applies an animation to an HTML element with specific trigger and oscillation settings.
+    /// - Parameters:
+    ///   - animation: The animation to apply to the element.
+    ///   - oscillate: Whether the animation should reverse when completed. Defaults to `true`.
+    ///   - trigger: The event that starts the animation (`.click`, `.hover`, or `.appear`).
+    /// - Returns: A modified copy of the HTML element with the animation applied.
+    func animation(_ animation: some Animation, oscillate: Bool = true, on trigger: AnimationTrigger) -> Self {
```
twostraws, line 32: For the sake of consistency, should `oscillate` here be `autoreverses`?

Sources/Ignite/Resources/js/animations.js
```diff
@@ -0,0 +1,80 @@
+function handleClickAnimation(event) {
```
twostraws, line 1: I think the chance of collision is low, but we're declaring various functions here and I don't want to accidentally overwrite other functions from other JavaScript code. Should these have some kind of `ignite` prefix or similar, to avoid that problem?

Sources/Ignite/Framework/Animations/Animatable.swift
```diff
@@ -0,0 +1,48 @@
...
+///
+///     var body: some Animation {
+///         StandardAnimation(property: "opacity", from: "0", to: "1")
+///     }
+/// }
+/// ```
+///
+/// - Note: This protocol is typically used in conjunction with the `Animation` protocol to create
+///         custom animations that can be applied to HTML elements.
+public protocol Animatable {
```
twostraws, line 29: Last night – maybe through lack of sleep! – the difference between `Animatable` and `Animation` was clear in my head. This  morning it is not 😅  I just tried removing `Animatable` and moving its properties into `Animation`, and it seemed to work equally well, removing a lot of the `if var copy = self as? any Animatable` and `return copy as! Self` stuff, and was less confusing for me. Am I missing something?

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+///         StandardAnimation(property: "transform", from: "scale(1)", to: "scale(1.1)")
+///             .duration(0.5)
+///             .timing(.easeInOut)
+///             .repeatCount(2, autoreverses: true)
+///     }
+/// }
+/// ```
+public protocol Animation {
+    associatedtype Body: Animation
+    var body: Body { get }
```
twostraws, line 29: This seems to return `self` everywhere it's used. Is it actually required, or can things like `buildBlock(component.body)` be written as `buildBlock(component)`?

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+            return copy as! Self
+        }
+        return self
+    }
+}
+
+public extension Animation {
+    @MainActor func register(for elementID: String) -> String {
+        if let resolvedAnimation = AnimationBuilder.buildBlock(self) as? ResolvedAnimation ??
+            AnimationBuilder.buildBlock(body) as? ResolvedAnimation
```
twostraws, line 118: This bit confused me regarding my previous question about `Animation.body` – if `body` always returns `self`, what's changing between the two sides of the nil coalesce?

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+    /// Creates a shaking animation.
+    static func shake() -> StandardAnimation {
+        StandardAnimation(property: "transform", from: "translateX(0)", to: "translateX(10px)")
+            .duration(0.5)
+            .timing(.easeInOut)
+            .repeatCount(3, autoreverses: true)
+    }
+    
+    /// Creates a 3D flip animation.
+    static func flip() -> StandardAnimation {
```
twostraws, line 224: It might be worth clarifying in what direction this is flipping.

Sources/Ignite/Framework/Animations/Animation.swift
```diff
@@ -0,0 +1,233 @@
...
+    /// Creates a bouncing animation.
+    static func bounce() -> StandardAnimation {
+        StandardAnimation(property: "transform", from: "translateY(0)", to: "translateY(-20px)")
+            .duration(0.5)
+            .timing(.custom("cubic-bezier(0.36, 0, 0.66, -0.56)"))
+            .repeatCount(1, autoreverses: true)
+    }
+    
+    /// Creates a shaking animation.
+    static func shake() -> StandardAnimation {
```
twostraws, line 216: If we wanted to keep close to SwiftUI we could match these names with their SF Symbols equivalents – `shake()` here appear to be "wiggle" in SF Symbols. I don't have a preference either way.

Sources/Ignite/Extensions/URL-SelectDirectories.swift
twostraws: @vdhamer Let's get this animation PR merged in, at which point you can check it and apply whatever you want on top. That should hopefully avoid too much extra to-ing and fro-ing. Thank you for your patience! 🙌

Sources/Ignite/Framework/Animations/KeyframeAnimation.swift
```diff
@@ -0,0 +1,64 @@
...
+///         StandardAnimation(property: "transform", from: "scale(1)", to: "scale(1.2)")
+///     }
+///     Frame("100%") {
+///         StandardAnimation(property: "transform", from: "scale(1.2)", to: "scale(1)")
+///     }
+/// }
+/// ```
+public struct KeyframeAnimation: Animation, Animatable {
+    /// The unique identifier for this keyframe animation
+    public let name: String
```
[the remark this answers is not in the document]
JPToroDev: You're right, we can drop the name. 🙌🏼

Sources/Ignite/Framework/Animations/Animation-Frame.swift
JPToroDev: Agreed. If we go with the closure approach, I don't think we'd need to expose the struct. So we could do something like `Animation.Frame` without making the user facing API messy.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Sure.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Sure.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Sure, I'll fix this.

Sources/Ignite/Framework/Animations/StandardAnimation.swift
JPToroDev: I think `BasicAnimation` works well—good call!

Sources/Ignite/Extensions/URL-SelectDirectories.swift
twostraws: I'm going to add this to a summary comment below, then mark this resolved.

Sources/Ignite/Framework/Animations/Animation.swift
twostraws: Great – thank you! 🙌

Sources/Ignite/Modifiers/Animation-Modifier.swift
JPToroDev: Yes, let's rename to `autoreverses`. We can set it to `false` by default.

Sources/Ignite/Resources/js/animations.js
JPToroDev: Good call, will prefix the JS functions 🙌🏼

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: As long as body is indeed returning `self` everywhere, we can definitely simplify!

Sources/Ignite/Framework/Animations/Animation-Frame.swift
twostraws: The closure approach is the best I've seen so far, particularly if you can get the builder/modifier tweak in there. I don't think `Animation.Frame` is possible because it would mean nesting a struct inside a protocol extension, but `KeyframeAnimation.Frame` certainly should be.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: I was likely mirroring another protocol's implementation here where `body` didn't always return `self`. It seems like here we don't need this check.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: Sure.

Sources/Ignite/Framework/Animations/Animation.swift
JPToroDev: I like it!

Sources/Ignite/Framework/Animations/Animation-Frame.swift
JPToroDev: Ah, of course—didn't register. Let's go with that then!

twostraws:
Now that we've gone through the code a second time, here's a fresh comment summing up the discussion so far.

From the first pass, you're going to:

- Add a commit to skip generating/including any animations code if there aren't any.
- Add a handful of extra comments here and there.
- Drop return in a handful of places.
- Investigate whether we can add animation triggers as an option set.
- Bring in your Property enum (presumably Animation.Property?), so we can remove some string-iness from the API.
- Change animation functions such as fadeIn() that have no parameters into computed properties.
- Switch to the same default timing and curves everywhere that's appropriate – bounce, shake, and flip should probably be excepted.
- Add some built-in spring types to mimic snappy and bouncy from SwiftUI, and also add an animation timing curve case that accepts four doubles for easier custom curves.
- Switch keyframe positions to numbers. SwiftUI doesn't have much (any?) prior art for percentages – a range of through through 1 as doubles is much more common. I'm not sure if it's worth adding overloads for an enum called KeyframeAnimation.Position that contains the cases from and to but internally just map to 0 and 1? (I suspect not.)

And we also suggested to push back to future work:

- Adding an anchor for various animations. If this happens in the future it will just be an extra parameter with a default value, so it shouldn't break anything.
- Whether CSS values in animations should have specific types such as .color(.red).


From the second pass, some of which might affect the above:

- When it's ready, we're going to merge this PR, and @vdhamer can add their own changes for building from within Xcode projects as one or more later PR.

- We've agreed to adjust the KeyframeAnimation API to the following:

```swift
KeyframeAnimation("crazy-colors")
    .keyframe(0) { content in
        content
            .backgroundColor(from: .blue, to: .yellow)
            .scale(from: 0, to: 1)
    }
    .keyframe(1) {
        $0.rotation(from: .degrees(0), to: .degrees(180))
    }
```

…although I'm still keen to think through what we name the animation methods in this place – `rotation()` vs `rotate()`, for example.

(If this ends up being tricky, we can push it back to a later PR.)

- You're going to switch over to Swift's `String.replacing()` rather than Foundation's `replacingOccurrences(of:)`.
- There are a handful of warnings to be addressed, e.g. `duration(copy.duration / speed)`.
- `StandardAnimation` to become `BasicAnimation`.
- User-facing animation names to be removed or made internal.
- `Frame` to become `KeyframeAnimation.Frame`
- `func animation(_ animation: some Animation, oscillate: Bool = true, on trigger: AnimationTrigger)` to become `func animation(_ animation: some Animation, autoreverses: Bool = false, on trigger: AnimationTrigger)`
- JavaScript functions to be prefixed with "ignite" or similar, to avoid the possibility of clashes.
- You're going to check whether `Animation.body` is indeed just using `self` everywhere, and potentially simplify to remove this requirement and associatedtype.
- The `flip()` animation will be renamed to clarify what direction it's flipping in. (Could easily add a second `flip…()` for the other direction too, or maybe make it a parameter?)
- You're going to look at – just look at, because it's only an idea! – the animation names from SF Symbols, to see if we can reach more alignment with them. If you hate them, please don't be afraid to ignore them entirely.


I've left open my question about Animatable and Animation, because I'm not sure where that should go.

Thank you so much! This is going to be an incredible new feature when it's merged in.

Sources/Ignite/Framework/Animations/StandardAnimation.swift
JPToroDev: The protocol enum constraint keeps slipping my mind. We can't do `Animation.Property`, but what we could do is have a property called `isAnimatable` in the regular `Property` enum and in the animation system use `assert` to cause an error if folks use a property that is not animatable. The alternative would be to create a separate `AnimatableProperty` enum with overlapping cases.

twostraws: Much as I dislike duplication, I think making the compiler enforce the rules here is a better idea, including for code completion reasons. That would mean having a second `AnimatableProperty` enum with overlapping cases. Is that okay?

Sources/Ignite/Framework/Animations/Animatable.swift
JPToroDev: So the reason I have both `Animation` and `Animatable` is so that we can do this:

```swift
struct MyNewAnimation: Animation {
    var body: some Animation {
        KeyframeAnimation()
            .frame(0) { content in
                content.backgroundColor("blue")
                content.color("yellow")
            }
            .frame(1) { content in
                content.backgroundColor("red")
                content.color("white")
            }
            .duration(10.0)
            .timing(.spring(dampingRatio: 0.5, velocity: 0.1))
    }
}
```

And not be prompted for `Animatable`'s properties. I can clean up the implementation a bit, though.

twostraws: Couldn't we make the properties internal, so folks just get prompted with the public methods?

JPToroDev: Sorry, I meant prompted for the methods.

twostraws: Could you just clarify a little more what you mean – where they will they be prompted?

JPToroDev: I moved the properties to `Animatable` because you can't have a proper setter in the default implementation of a protocol and you can't set the access level of properties. The system animations—`BasicAnimation`, `KeyframeAnimation`, etc.— have their own implementations of the `Animatable` properties so that they can be set and get. If I combined `Animatable` and `Animation`, folks would be prompted to add these properties to their own animations whether used or not. They aren't really relevant to custom animations and would muddied a clean API.

Sources/Ignite/Framework/Animations/AnimationClassGenerator.swift
JPToroDev: So some of these `return` keywords are necessary to avoid compiler errors. This is one case.

Sources/Ignite/Framework/Animations/AnimationCurve.swift
JPToroDev: These ones can be safely removed.

JPToroDev:
This builds and runs fine through the command line, but not through Xcode. This isn't an issue in my private fork, and despite comparing both versions of the framework several times, I still haven't pinpointed the issue. All to say, this should resolve itself once all the PRs are merged, and in the meantime I'll continue to bug hunt.

Sources/Ignite/Framework/Animations/Angle.swift
```diff
@@ -0,0 +1,27 @@
+//
+// Angle.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+/// Represents an angle that can be expressed in different units
+public enum Angle: Sendable {
```
twostraws, line 9: Love this!

Sources/Ignite/Framework/Animations/Rotation.swift
```diff
@@ -0,0 +1,21 @@
...
+// Rotation.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+/// Defines the possible directions for 3D rotations
+public enum Rotation: Sendable {
+    /// Rotates 360 degrees around the Y axis (left to right)
+    case forward
```
twostraws, line 11: To avoid complexities around culture differences of "forward", I think maybe just "left" and "right" here?

vdhamer:
I did a test integration of current main branch with this PR #143 applied.
Ran into breaking changes in my code (which is ok), but am looking for documentation or an examples to make `Theme` work again. Incomplete list of issues, to give you guys an idea of what's happened:

- Theme.render() is gone and replaced by Theme.body
- HTML complains about `any`
- the typing of the function elements within Body is changed
- compiler can't find `Component` and `PageElement` anymore (type change)

I realize that in theory I can figure out the changes for users from the code. But is there an update to the README coming?

```swift
struct MyTheme: Theme {
    func render(page: Page, context: PublishingContext) -> HTML {
        HTML {
            Head(for: page, in: context)

            Body {
                page.body

                IgniteFooter() // "Created with Ignite"
                UpdatedOnFooter() // my own code
            }
        }
    }
}
```

twostraws:
We'll be getting the current big PR merged shortly, I think, at which point there's hopefully only one or two more. As soon as that's done my number one priority will be updating the documentation and all sample code.

vdhamer:
👍

Sources/Ignite/Framework/Animations/Rotation.swift
JPToroDev: You got it.

Sources/Ignite/Framework/Animations/StandardAnimation.swift
JPToroDev: What do you think about renaming these to `SingleStepAnimation` and `MultiStepAnimation`?

twostraws:
Are you happy for me to merge this?

JPToroDev:
Ha! Got this message right as I hit send on an email to you. All good—FINALLY!

twostraws:
Thank you! I'll merge it now, and we can make a plan for next steps.

PR #146, opened 2024-11-15
Fix URL init compiler error

denis-rybkin:
Compiler error when initializing the URL, which occurs because the project has an URL extension(file URL-Unwrapped.swift) that conflicts with Foundation’s initializer signature.
Screenshot 2024-11-15 at 3 14 33 AM
Fixed extension, files using it and tests. Sample project will be fixed after merge.
This error can also be fixed by force-unwrapping the URL. Since the project has no runtime, and if the url is incorrect, a crash will occur during the build. In this case extension will be deleted.


twostraws:
Thank you!

PR #149, opened 2024-11-29
Introduce a Revamped `Theme` Protocol and `Font` Struct

JPToroDev:
# Theme System and Typography Enhancements

> **Note:** 
> - The type previously named `Theme` has been renamed to `Layout` in this update
> - Several theme-related properties in `Site` have been renamed accordingly

This PR introduces a comprehensive theming system and typography management for Ignite websites. The changes provide full control over colors, fonts, and responsive design while maintaining compatibility with Bootstrap's theming approach.

### Theme Management
- Introduced a new `Theme` protocol for defining complete website themes with flexible configuration options
- Implemented three built-in themes:
 - `LightTheme`: Bootstrap's default light theme
 - `DarkTheme`: Bootstrap's dark theme variant
 - `AutoTheme`: Automatically switches between light/dark based on system preferences
- Added automatic theme switching support via the new `SwitchTheme` action
- Implemented CSS variable generation for all theme properties to ensure consistent styling

### Font System
- Consolidated `CustomFont` and `Font` into a revamped type for defining custom fonts and font stacks

### Site Configuration
```swift
struct ExampleSite: Site {
   var name = "Hello World"
   var titleSuffix = " – My Awesome Site"
   var url = URL(static: "https://www.example.com")
   var builtInIconsEnabled = true
   var author = "John Appleseed"
   var homePage = Home()
   var layout = MyLayout()
   var darkTheme: any Theme = MyDarkTheme()
   var lightTheme: any Theme = MyLightTheme()
   var alternateThemes: [any Theme] = [MyCustomTheme()]
}
```

### Custom Layout Definition
```swift
struct MyLayout: Layout {
   var id: String = "my-theme"
   var largeMaxWidth: any LengthUnit = "900px"
   var xxLargeMaxWidth: any LengthUnit = "900px"
   var font = Font(name: "valkyrie_a_regular", source: "/fonts/valkyrie_a_regular.woff2")
}
```

### Example Usage: Creating a Theme Switcher
With the new theme system, you can easily create interactive components like a theme switcher. Here's an example of how you could implement one:

```swift
struct ThemeSwitcher: HTML {
   @Environment(\.themes) var themes
   var body: some HTML {
        ForEach(themes) { theme in
            Button(theme.id) {
	        SwitchTheme(theme.id)
	    }
       }
   }
}
```

## Breaking Changes
- The previous `Theme` protocol has been renamed to `Layout`
- Several properties in `Site` have been renamed accordingly

## In Progress
- [ ] Add unit tests for new theme/layout functionality

## Migration Guide
To upgrade existing code:
1. Rename all instances of `Theme` protocol to `Layout`
2. Update theme-related properties in `Site` to reflect multi-theme support

## Dependencies
- No new dependencies added

Let me know if you have any questions!


Sources/Ignite/Framework/Environment/EnvironmentValues.swift
```diff
@@ -49,6 +53,7 @@ public struct EnvironmentValues: Sendable {
         self.allContent = allContent
         self.feedConfiguration = site.feedConfiguration
         self.isFeedEnabled = site.isFeedEnabled
+        self.themes = [site.lightTheme, site.darkTheme, AutoTheme()] + site.alternateThemes
```
twostraws, line 56: I know this is the default behavior from Bootstrap, but I want to clarify it here. Should sites automatically display light or dark mode based on the system settings? I think so, but I'm keen to hear any alternative views. That doesn't mean it should be _required_ to have both modes, and in fact making it something that can be disabled is a good idea, however it would match SwiftUI to make the sensible behavior the default.

Sources/Ignite/Framework/Layout.swift
```diff
@@ -23,7 +23,7 @@ import Foundation
 /// }
 /// ```
 @MainActor
-public protocol Theme: Sendable {
+public protocol Layout: Sendable {
```
twostraws, line 26: Ignite was rather wooly about "theme", "layout", and "content page", which is my fault – I set out on the direction I wanted, then kind of did some course corrections based on Jekyll spellings, but it should have been clearer all along. This PR goes some way to clearing up that confusion, but just for my own understanding:

- A `Theme` controls the coloring, font selection, etc.
- A `Layout` is a structure for rendering a page, which might be a static page or a content page.
- A `ContentPage` is something that has some Markdown attached, which adds extra HTML to a `Layout`.
- A `StaticPage` is something that is pure Swift.

Is that accurate? If so, would `ContentLayout` and `StaticLayout` work better? Is `ThemePage` still needed?

You can see an example of the Jekyll that directed me here: https://github.com/swiftlang/swift-org-website/blob/main/_layouts/post.html

Sources/Ignite/Framework/Site.swift
```diff
@@ -49,10 +49,6 @@ public protocol Site: Sendable {
     /// The base URL for your site, e.g. https://www.example.com
     var url: URL { get }
 
-    /// How wide your page should be on desktop web browsers. Defaults to 10,
-    /// which allows a little margin on either side.
-    var pageWidth: Int { get }
```
twostraws, line 54: Where did this move to?

Sources/Ignite/Framework/Site.swift
```diff
@@ -89,7 +85,19 @@ public protocol Site: Sendable {
...
+    var lightTheme: any Theme { get }
+
+    /// The theme used when the system is in dark mode.
+    var darkTheme: any Theme { get }
+
+    /// Additional themes that can be selected by users beyond light and dark mode.
+    var alternateThemes: [any Theme] { get }
+
+    /// JavaScript code for analytics tracking, injected into the page head.
+    var analyticsSnippet: Script? { get }
```
twostraws, line 100: Please remove this; we should return to it in the future to look at something better than a string – see YouTube and Vimeo embeds, for example.

Sources/Ignite/Publishing/PublishingContext.swift
```diff
@@ -177,6 +172,7 @@ public final class PublishingContext {
         }
 
         try copy(resource: "js/email-protection.js")
+        try copy(resource: "js/theme-switcher.js")
```
twostraws, line 175: Unless these extra pieces of JavaScript need to be separate, I think we should bundle them into a single JavaScript file. That could be done statically right in the Ignite repository, or we could ship them separate and join them at build time, I don't mind. If we do it statically, it would mean adding a couple of line breaks plus a comment between each discrete piece of functionality.

Sources/Ignite/Publishing/PublishingContext.swift
```diff
@@ -344,7 +386,7 @@ public final class PublishingContext {
     /// - Returns: The correct `ContentPage` instance to use for this content.
     func layout(for content: Content) throws -> any ContentPage {
         if let contentLayout = content.layout {
-            for layout in site.layouts {
+            for layout in site.contentPages {
```
twostraws, line 389: layouts in contentPages doesn't seem right – again, one of these two names should change.

Sources/Ignite/Themes/DarkTheme.swift
```diff
@@ -0,0 +1,52 @@
...
+/// Conforming types automatically inherit all default dark theme values and can override specific properties as needed.
+///
+/// Example:
+/// ```swift
+/// struct CustomDarkTheme: DarkTheme {
+///     var id: String = "custom-dark"
+///     var primary: String = "#990000" // Override just the primary color
+/// }
+/// ```
+public protocol DarkTheme: Theme {}
```
twostraws, line 20: This seems like a concrete implementation of a theme – why not a struct?

Sources/Ignite/Themes/Fonts/FontSource.swift
```diff
@@ -0,0 +1,85 @@
...
+///
+/// ```swift
+/// // Local font
+/// var fontFamilyBase = Font(
+///     name: "valkyrie_a_regular",
+///     source: "/fonts/valkyrie_a_regular.woff2"
+/// )
+///
+/// // Remote font
+/// var fontFamilyBase = Font(
```
twostraws, line 22: AT A LATER DATE: I know Google Fonts are very popular, so if the URL format is really this fixed, a non-stringly-typed initializer would be lovely – something like a weights array of integers and an array of font styles (`.normal`, `.italic`, etc) *might* work well. (Not for this PR though, please.)

Sources/Ignite/Themes/Theme-DefaultImplementation.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    // Font Families
+    var sansSerifFont: Font { .systemSansSerif }
+    var monospaceFont: Font { .systemMonospace }
+    var font: Font { .systemBodyFont }
+    var codeFont: Font { .systemCodeFont }
+    var alternateFonts: [Font] { [] }
+
+    // Font Sizes
+    var rootFontSize: (any LengthUnit)? { nil } // null by default to use browser default
```
twostraws, line 40: There are so many variants of things in this part of the code – xxLarge, xLarge, large, etc. While I appreciate they should all be individually customizable, would it make sense / be possible to add some kind of base + multiplier approach? This would mean:

1. Someone can individually customize every single option for sizes.
2. Someone could say "my base breakpoint size / heading size / page width is X, figure out the rest for me based on that" – and we'd apply default (overridable) multipliers for each of the variants from that.
3. Someone could say "my base breakpoint size is X, figure out the rest for me, but after you've done that I want to change `xxLargeBreakpoint` to be something custom."

I don't know if it's possible or even makes sense; I'm just looking for ways to make the ergonomics a little easier.

Sources/Ignite/Framework/MissingLayout.swift
```diff
@@ -9,8 +9,8 @@ import Foundation
 
 /// A theme that does nothing at all. This is used as the default theme, so we can
```
twostraws, line 10: "Layout"?

Sources/Ignite/Publishing/PublishingContext.swift
```diff
@@ -227,11 +223,11 @@ public final class PublishingContext {
 
     /// Renders one page using the correct theme, which is taken either from the
     /// provided them or from the main site theme.
-    func render(_ page: Page, using theme: any Theme) -> String {
+    func render(_ page: Page, using theme: any Layout) -> String {
```
twostraws, line 226: `theme: any Layout` hurts my head here.

Sources/Ignite/Themes/Theme.swift
```diff
@@ -0,0 +1,198 @@
...
+    var danger: Color { get }
+
+    /// Light theme color
+    var light: Color { get }
+
+    /// Dark theme color
+    var dark: Color { get }
+
+    /// Default text color for body content
+    var primary: Color { get }
```
twostraws, line 53: Is this the default text color?

Sources/Ignite/Themes/Theme.swift
```diff
@@ -0,0 +1,198 @@
...
+///     var primary: String = "#ff0000"
+///     var fontFamilyBase: Font = .custom("Helvetica")
+/// }
+/// ```
+public protocol Theme: Sendable {
+    /// Unique identifier for the theme, automatically sanitized to lowercase with hyphens
+    var id: String { get set }
+
+    /// Primary brand color
+    var accent: Color { get }
```
twostraws, line 29: Where would this be used in practice? An example would be helpful.

Sources/Ignite/Themes/Theme.swift
```diff
@@ -0,0 +1,198 @@
...
+    var secondary: Color { get }
+
+    /// Color for tertiary content
+    var tertiary: Color { get }
+
+    /// Default background color
+    var primaryBackground: Color { get }
+
+    /// Secondary background color
+    var secondaryBackground: Color { get }
```
twostraws, line 68: Some examples of where this is used would be helpful. For example, does it affect the accordion open background color, or is that something else?

Sources/Ignite/Themes/DarkTheme.swift
twostraws: Having read further, I think this is because it's designed so users can build on `DarkTheme` to make small tweaks?

twostraws:
Thank you for this! I hope this is the last significant PR that's coming, because we desperately need to get this thing stable 🙂

I've added a bunch of questions and comments, but the TL;DR is: I love it! Lots of scope for customization to really give developers the chance to make sites theirs. The broad areas of concern I have are:

- There's still some confusing wording around Theme/ContentPage/Layout. Your example Custom Theme Definition above says `struct MyTheme: Layout {`, which I just don't understand. I'd like to get all this wording cleared up and finalized everywhere, so we're really tight.
- If we're going to do analytics, it needs to be a separate PR where we can think more about common services – it might be the case that we make some specific types to handle common services like GA.
- I want to avoid a proliferation of JavaScript/CSS files if possible. I know with keepalive the actual cost is limited, but it just gets kinda messy. I don't mind shipping lots of files separately with Ignite, but we should merge them at build time, or find some other solution.
- It would be helpful if we could document how various theme colors are used. Some are self-evident like hover color, but if I wanted to change the background color of an open accordion item, where would that be? Does changing the `danger` theme color change the text and the background to the same thing, automatically change one to a variant of the other, or change only one?
- If I wanted all the fonts to be a little bigger, should I change Theme.font? Theme.rootFontSize, Theme.bodySize, or all the font size variants? Does one override the other? Would using .font(.title2) automatically adapt based on these?

Sources/Ignite/Framework/Environment/EnvironmentValues.swift
JPToroDev: > Should sites automatically display light or dark mode based on the system settings?

That's correct. If you'd like to give users the option to disable automatically using the system's color scheme, we could: 1) make `lightTheme` and `darkTheme` optional, and if they want to disable one they can make it `nil`. If they mark both `nil`, we could throw a compile-time error. Or 2), we leaves things as they are and create a bool to disable automatic switching. I'd be leaning toward the first.

Sources/Ignite/Framework/Layout.swift
JPToroDev: Yes, that's correct. 

>  If so, would ContentLayout and StaticLayout work better?
 
Ooo, I like that! We could also use the "template" instead of "layout" throughout this PR. Is that preferable to "layout"?
 

>  Is ThemePage still needed?
 
 This is an excellent point. I don't think we need this type anymore.

Sources/Ignite/Framework/Site.swift
JPToroDev: This is handled by `Theme` now.

From my example above:

```swift
var largeMaxWidth: any LengthUnit = "900px"
var xxLargeMaxWidth: any LengthUnit = "900px"
```

Sources/Ignite/Framework/Site.swift
JPToroDev: Sure.

Sources/Ignite/Framework/ThemedPage.swift
```diff
@@ -12,7 +12,7 @@ import Foundation
 @MainActor
 public protocol ThemePage: EnvironmentReader {
     /// The type of theme you want this page to use.
-    associatedtype ThemeType: Theme
+    associatedtype ThemeType: Layout
```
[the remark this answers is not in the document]
JPToroDev: I missed including the updated version of this file in the PR. All instances of "theme" should be replaced with "layout".

Sources/Ignite/Publishing/PublishingContext.swift
JPToroDev: Sure, let's do it statically.

Sources/Ignite/Publishing/PublishingContext.swift
JPToroDev: It should be:

```swift
for page in contentPages {
```

Will fix.

Sources/Ignite/Themes/DarkTheme.swift
JPToroDev: Exactly.

Sources/Ignite/Themes/Fonts/FontSource.swift
JPToroDev: I agree—there was something I encountered that made me think that might not always be the case, but I'll have to do a proper deep dive down the line.

Sources/Ignite/Themes/Theme-DefaultImplementation.swift
JPToroDev: I like that idea, and it's possible—we'd simply make the multipliers optional, and if a multiplier exists, strip any length units at the end of the base string, multiply the numerical value, then append those units to the new value to create a new string.

It diverges from how Bootstrap does things, which isn't alway the most ergonomic, but that pattern does seem nice. My only concern: things get messy if somebody sets a base setting and multiplier, but then sets non-base variants too. Would those explicit variant settings simply override the multiplier values? Would mixing these two systems potentially confuse people?

Sources/Ignite/Framework/MissingLayout.swift
JPToroDev: Yep, will update.

Sources/Ignite/Publishing/PublishingContext.swift
JPToroDev: Unintentional, will fix.

Sources/Ignite/Themes/LightTheme.swift
```diff
@@ -0,0 +1,31 @@
...
+/// Conforming types automatically inherit all default light theme values and can override specific properties as needed.
+///
+/// Example:
+/// ```swift
+/// struct CustomLightTheme: LightTheme {
+///     var id: String = "custom-light"
+///     var primary: String = "#ff0000" // Override just the primary color
+/// }
+/// ```
+public protocol LightTheme: Theme {}
```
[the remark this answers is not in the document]
JPToroDev: Yessir.

Sources/Ignite/Themes/Theme.swift
JPToroDev: Yes it is.

Sources/Ignite/Themes/Theme.swift
JPToroDev: Things like accordion background color are determined by their own classes and variables. In order to change details like that, we'd have to override those classes / variables. I think that'd best be handled by a modifier. I can make one for accordion as an example in my next commit if you'd like.

twostraws: Please don't; let's get this merged in, then we can look for ways to improve incrementally. I just had _another_ email about someone trying and failing to use Ignite, so we're well into "burning goodwill" territory now.

Sources/Ignite/Framework/Environment/EnvironmentValues.swift
twostraws: I see a handful of situations:

1. If they have light and dark theme enabled, they might want automatic switching based on system settings. This is the SwiftUI approach.
2. They disable *either* light mode or dark mode, in which case the site always runs in one mode.
3. They enable light and dark mode, but want the user to be able to switch, to accommodate users who find reading dark mode websites difficult.

JPToroDev:
This all sounds great, Paul!

> Does changing the danger theme color change the text and the background to the same thing, automatically change one to a variant of the other, or change only one?

Could you elaborate on this a little more? It doesn't look like `danger` has text and background base variants according to the Bootstrap documentation.

> If I wanted all the fonts to be a little bigger, should I change Theme.font? Theme.rootFontSize, Theme.bodySize, or all the font size variants?

`rootFontSize` would do the trick. I think the default is `16px` in most browsers. But there's actually a bug preventing this from working that I'll fix in the next commit.

Sources/Ignite/Framework/Layout.swift
twostraws: Layout follows Jekyll, which is nice. At some point I'd also like to bring in an optional templating engine such as Mustache, so we need to be careful around the word "template".

Sources/Ignite/Framework/Site.swift
twostraws: Makes sense.

Sources/Ignite/Publishing/PublishingContext.swift
twostraws: Great!

Sources/Ignite/Themes/Fonts/FontSource.swift
twostraws: Okay; one for the issues list to look at later.

twostraws:
> Could you elaborate on this a little more? It doesn't look like danger has text and background base variants according to the Bootstrap documentation.

If I set my theme's danger color to #FF0000, is that changing the danger text color to red, the danger background color to red, or something else?

Sources/Ignite/Framework/Environment/EnvironmentValues.swift
JPToroDev: OK, then let's do with the first option I outlined. It's more dynamic. Developers can add a theme switcher like the one I included as an example above.

Sources/Ignite/Framework/Layout.swift
JPToroDev: OK, we'll stick with "layout". 🙌🏼

JPToroDev:
> > Could you elaborate on this a little more? It doesn't look like danger has text and background base variants according to the Bootstrap documentation.
> 
> If I set my theme's danger color to #FF0000, is that changing the danger text color to red, the danger background color to red, or something else?

It'd change the color anywhere the Bootstrap `danger` CSS variable is used. So if that's used in text, it'd change there, and if that's used in a background, it'd change there, too.

Sources/Ignite/Framework/Environment/EnvironmentValues.swift
twostraws: Sounds good to me!

Sources/Ignite/Framework/Layout.swift
twostraws: Perfect 👍

Sources/Ignite/Themes/Theme-DefaultImplementation.swift
twostraws: Let's leave this one as-is and come back to it later on after it's had more discussion.

Sources/Ignite/Themes/Theme.swift
```diff
@@ -0,0 +1,198 @@
...
+    var emphasis: Color { get }
+
+    /// Color for secondary content
+    var secondary: Color { get }
+
+    /// Color for tertiary content
+    var tertiary: Color { get }
+
+    /// Default background color
+    var primaryBackground: Color { get }
```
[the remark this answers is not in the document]
twostraws: Great!

Sources/Ignite/Themes/Theme.swift
twostraws: Wait, so what's `primary` then? 😅

twostraws:
> It'd change the color anywhere the Bootstrap danger CSS variable is used. So if that's used in text, it'd change there, and if that's used in a background, it'd change there, too.

Well, let's give it a try and see how we get on!

Once the changes discussed above are ready, we can merge this and move on.

JPToroDev:
OK, great. Should have this all ready to go within the hour.

Sources/Ignite/Themes/Theme.swift
JPToroDev: `primary` maps to the color for body text, like SwiftUI.

twostraws: Ah! Yes, that makes sense from our API's perspective.

Sources/Ignite/Themes/Theme-DefaultImplementation.swift
JPToroDev: Ok great!

JPToroDev:
All ready to go. Let me know if I missed anything.

Sources/Ignite/Framework/TagPage.swift
```diff
@@ -21,7 +21,8 @@ import Foundation
 ///     }
 /// }
 /// ```
-public protocol TagPage: ThemePage {
+@MainActor
+public protocol TagPage: EnvironmentReader {
```
twostraws, line 25: TagLayout?

Sources/Ignite/Themes/Fonts/BootstrapVariable.swift
```diff
@@ -0,0 +1,183 @@
...
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+/// A collection of CSS variables used by Bootstrap for theming.
+///
+/// Bootstrap variables control various aspects of a theme including colors, typography,
+/// and spacing. Each case represents a specific CSS variable that
+/// Bootstrap uses for consistent styling across components.
+public enum BootstrapVariable: String {
```
twostraws, line 13: Does this need to be public? It gets really confusing around `primary`, `bodyColor`, etc.

twostraws:
I've added two small queries, but otherwise looking great!

Sources/Ignite/Framework/TagPage.swift
JPToroDev: Good catch. Updated!

Sources/Ignite/Themes/Fonts/BootstrapVariable.swift
JPToroDev: Another good catch—updated!

twostraws:
Is this good to merge, or are you still tweaking?

JPToroDev:
Squeaky clean now. All set.

twostraws:
Thank you!

PR #152, opened 2024-12-04
Add theme-specific syntax-highlighting color schemes...

JPToroDev:
Add theme-specific syntax-highlighting color schemes. Reduce syntax-highlighting overhead by moving from dormant Prism to active Highlight.js.


Sources/Ignite/Elements/MetaLink.swift
```diff
@@ -28,15 +28,31 @@ public struct MetaLink: HeadElement, Sendable {
...
     /// The CSS responsible for controlling the visibility of environment-dependent elements.
     static let customCSS = MetaLink(href: "/css/custom.min.css", rel: .stylesheet)
     static let utilityCSS = MetaLink(href: "/css/utilities.min.css", rel: .stylesheet)
     static let themeCSS = MetaLink(href: "/css/themes.min.css", rel: .stylesheet)
     static let animationCSS = MetaLink(href: "/css/animations.min.css", rel: .stylesheet)
 
+    /// Creates an array of `MetaLink` elements for syntax highlighting themes.
+    /// - Parameter themes: A collection of syntax highlighting themes to include.
+    /// - Returns: An array of MetaLink elements. If multiple themes are provided, includes data attributes for theme switching.
+    static func highlighThemeMetaLinks(for themes: some Collection<HighlighterTheme>) -> [MetaLink] {
```
twostraws, line 40: Nit: "highlighThemeMetaLinks" is missing a T.

Sources/Ignite/Elements/MetaLink.swift
```diff
@@ -28,15 +28,31 @@ public struct MetaLink: HeadElement, Sendable {
...
+        let hasMultipleThemes = themes.count > 1
+
+        return themes.map { theme in
```
twostraws, line 43: Is this going to create a whole bunch of CSS files to be imported?

twostraws:
I'm broadly fine with the move to a different library for highlighting, but I'm dead against loading remote JS/CSS – I'd rather we had a local clone of it all inside Ignite, bundling those files together as needed. This would allow our sites to work offline, including for folks who embed in apps.

Sources/Ignite/Elements/MetaLink.swift
JPToroDev: Great catch—thank you!

Sources/Ignite/Elements/MetaLink.swift
JPToroDev: Only the themes in use by the site.

JPToroDev:
This should be good to go now!

Sources/Ignite/Elements/MetaLink.swift
twostraws: In the future we should perhaps look to coalesce them at build time.

twostraws:
Thank you!

PR #154, opened 2024-12-04
Minor improvements to `List` API...

JPToroDev:
Minor improvements to `List` API. Add `group` and `flushGroup` list styles.


Sources/Ignite/Framework/ElementTypes/HTML.swift
```diff
@@ -97,7 +97,7 @@ public extension HTML {
     /// Adds a single CSS class to the element.
     /// - Parameter className: The name of the CSS class to add
     /// - Returns: The modified HTML element
-    func `class`(_ className: String) -> Self {
+    @discardableResult func `class`(_ className: String) -> Self {
```
twostraws, line 100: It's discardable here too? You must really want this to be discardable 😅

JPToroDev: I made that discardable to silence a compiler warning. I could make a different version of the method that doesn't return anything—would that be preferable?

JPToroDev: It was 'The result of this method isn't being used', within this block I'm pretty sure:

```swift
 func body(content: some HTML) -> any HTML {
        if content.body.isComposite {
            if let customHeight {
                content.containerStyle(.init(name: "line-height", value: String(customHeight)))
            } else if let presetHeight {
                content.containerClass("lh-\(presetHeight.rawValue)")
            }
        } else {
            if let customHeight {
                content.style("line-height: \(customHeight)")
            } else if let presetHeight {
                content.class("lh-\(presetHeight.rawValue)")
            }
        }
        content
    }
```

twostraws: What is the `content` doing at the end there? That seems dubious.

JPToroDev: Inside `List`'s rendering method:

```swift
for item in items {
            // Any element that renders its own <li> (e.g. ForEach) should
            // be allowed to handle that itself.
            if let listableItem = item as? ListableElement {
                if listStyle != .plain {
                    item.class("list-group-item") /* <-- Here */
                }
                output += listableItem.renderInList(context: context)
            } else {
                let styleClass = listStyle != .plain ? " class=\"list-group-item\"" : ""
                item.class("m-0")  /* <-- Here */
                output += "<li\(styleClass)>\(item.render(context: context))</li>"
            }
        }
```

twostraws: I see. I think perhaps returning `Self` is conflicting with the global attribute store ID system. Leave it alone for now, and we can circle back later.

JPToroDev: I think this stems from uses like this:

```swift
// Position modifier
func body(content: some HTML) -> any HTML {
        content.class(position.rawValue)
}
```
Where the method updates `AttributeStore` and acts as a passthrough for the view. In `List`, we don't need the passthrough element, just store updating element. Perhaps it's a little clunky, but I don't think it's dangerous.

JPToroDev: By leave the warning as is, do you mean remove `@discardableResult`? If not, this should be good to go.

twostraws: Yeah, let's remove `@discardableResult` for now, if that's okay; the warning is a tiny nag to resolve it fully in the future.

JPToroDev: You got it. All set.

twostraws:
Thank you!

PR #155, opened 2024-12-04
Integrate SwiftSoup for HTML prettification.

JPToroDev:


twostraws:
I'm not a big fan of adding dependencies in general – they add download size, affect compile time locally and through CI, and add an ongoing reliance on something out of our control. However, this package is used in only a very small way, so if worse comes to worst we can take it out trivially in the future, so I'm happy to merge it in here. Thank you!

PR #156, opened 2024-12-04
Add `lineHeight()` modifier. Improve uniformity of HTML method names.

JPToroDev:


Sources/Ignite/Framework/ElementTypes/HTML.swift
```diff
@@ -97,7 +97,7 @@ public extension HTML {
     /// Adds a single CSS class to the element.
     /// - Parameter className: The name of the CSS class to add
     /// - Returns: The modified HTML element
-    func `class`(_ className: String) -> Self {
+    @discardableResult func `class`(_ className: String) -> Self {
```
twostraws, line 100: Why is this now discardable?

Sources/Ignite/Modifiers/LineHeight.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+/// Predefined line height values that match Bootstrap's spacing system.
+public enum LineHeight: String, CaseIterable {
+    /// Single line height (1.0)
+    case tight = "1"
+
+    /// Compact line height (1.25)
+    case small = "sm"
```
twostraws, line 84: I'm not a massive fan of cases where two words seem interchangeable – `tight` and `small` here feel like they could be reversed. This happens in SwiftUI too, where I usually look up whether `light` or `thin` is lighter! So, how about large, default, small, and xSmall, or something similar?

Sources/Ignite/Modifiers/LineHeight.swift
```diff
@@ -0,0 +1,101 @@
...
+    func lineHeight(_ height: LineHeight) -> some HTML {
+        modifier(LineHeightModifier(height: height))
+    }
+}
+
+public extension BlockHTML {
+    /// Sets the line height of the element using a custom value.
+    /// - Parameter height: The line height multiplier to use
+    /// - Returns: The modified HTML element
+    func lineHeight(_ height: Double) -> some BlockHTML {
```
twostraws, line 66: SwiftUI calls this `lineSpacing()`, but I'm not sure whether it refers to precisely the same thing or not.

Sources/Ignite/Framework/ElementTypes/HTML.swift
JPToroDev: It silenced a warning the compiler was giving in the `lineHeight()` modifier struct. We could create an alternate version of the method that doesn't return anything?

Sources/Ignite/Modifiers/LineHeight.swift
JPToroDev: I think it does the same thing—I'll rename.

Sources/Ignite/Modifiers/LineHeight.swift
JPToroDev: You got it.

Sources/Ignite/Modifiers/LineHeight.swift
twostraws: I don't know precisely what it maps to in HTML/CSS, but Apple has often provided varying options here. See [this](https://developer.apple.com/documentation/appkit/nsparagraphstyle/lineheightmultiple) and [this](https://developer.apple.com/documentation/appkit/nsparagraphstyle/linespacing) in `NSParagraphStyle`, for example.

JPToroDev: It maps to CSS's `line-height` property. I think we could go either way as long as the documentation is clear.

twostraws: Okay; let's stick with your latest version for now; we can revisit later if it needs more honing.

JPToroDev: OK, then this is good to merge.

twostraws:
Thank you!

PR #161, opened 2024-12-07
Make FeedLink public.

gavinjerman:
Make the FeedLink component public so that it may be used by Ignite clients.


twostraws:
Thank you!

PR #165, opened 2024-12-11
Add support for Markdown hard line breaks

gavinjerman:
Ensure a line break in Markdown (not in a code span or HTML tag) that is preceded by two or more spaces and does not occur at the end of a block is parsed as a hard line break (rendered in HTML as a <br /> tag).


twostraws:
Thank you!

PR #167, opened 2024-12-12
Implement `colorWeight()` modifier…

JPToroDev:
Implement `colorWeight()` modifier. Rework `Color` and scope HTML colors.


Sources/Ignite/Styles/Color-HTMLColors.swift
```diff
@@ -0,0 +1,454 @@
...
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+// There are lots of HTML color names here!
+public extension Color {
+    /// A namespace containing standard HTML/CSS color names and their hex values.
+    /// These colors are based on the standard web colors specification.
+    enum HTML {
```
twostraws, line 12: How does pushing colors one level deeper affect the API – can users still say `foregroundStyle(.rebeccaPurple)`, for example?

JPToroDev: Users would now do: `foregroundStyle(.HTML.rebeccaPurple)`. HTML colors had to be scoped because several HTML color names conflict with Bootstrap color names.

twostraws: I'm reading this again: if there are collisions, HTML needs to win, so we should remove the colliding Bootstrap color names. I'd rather avoid `.HTML.rebeccaPurple` if possible.

JPToroDev: So for Bootstrap colors, should we do boostrapBlue, etc.?

twostraws: Yes, that sounds good.

JPToroDev: Great—all set then.

JPToroDev: Actually, on second thought: given the future direction of the framework we discussed, I think we should revert `Color` back to a struct and add `colorWeight()` to that. Hold tight.

JPToroDev: All set.

PR #172, opened 2024-12-12
Introduce form components to create simple newsletter signups.

JPToroDev:
This PR introduces three form components—`Form`, `TextField`, and `SubscribeAction`—that allow users to craft basic email intake forms and create a foundation for future, full-fledged form creation.

Usage example:

```swift
Form {
    TextField(placeholder: "Email")
    Button("Submit")
        .type(.submit)
    } onSubmit: { id in
        SubscribeAction(.sendFox("myToken"), formID: id)
}
```


Sources/Ignite/Actions/SubscribeAction.swift
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /// The endpoint URL for the form submission
+    var endpoint: String {
+        service.endpoint
+    }
+
+    /// Returns the JavaScript needed for form handling
+    public func compile() -> String {
+        """
```
twostraws, line 65: This whole PR is excellent – great job! This is the only part I'm not sure about: what JavaScript is required for the subscribe action?

JPToroDev: Thank you! Good catch—the JS just includes the form reset logic here. I'll update this with the subscription-handling logic that was omitted.

Sources/Ignite/Framework/ElementTypes/BlockHTML.swift
```diff
@@ -24,11 +24,6 @@ extension BlockHTML {
     public func width(_ width: Int) -> Self {
         var copy = self
         copy.columnWidth = .count(width)
-
-        // Important: We need to transfer all our existing
-        // attributes over to the copy.
-        AttributeStore.default.merge(attributes, intoHTML: copy.id)
```
twostraws, line 30: I was reading through again and noticed this change – can this safely go? I seem to recall it causing problems in Ignite Samples.

JPToroDev: The copy should have the same id as the original, so this shouldn't be necessary.

twostraws: Great – thanks!

PR #179, opened 2024-12-20
Replace #file with #filePath in IgniteTests.swift

erikbasargin:
This PR addresses the `Ignite.PublishingError.missingPackageDirectory` error that occurred in `IgniteTests.swift` when using `#file` by replacing it with `#filePath`.


twostraws:
Thank you!

PR #180, opened 2024-12-20
Fix font application

erikbasargin:
This issue is detected by the `test_customFont` test when the #179 PR is applied. By modifying an underlying HTML content within the `ModifiedHTML`, the modifier `FontStyleModifier` can find the text content and apply required style.

These changes are proposed based on the assumption that `ModifiedHTML` need to modify underlying HTML content rather than itself (`FontStyleModifier` expects `content.body` to be `Text`). However, I can be mistaken because it contradicts to the merged #174 changes. Please let me know what you think.

### DIFF
--- a/Sources/Ignite/Modifiers/ModifiedHTML.swift
+++ b/Sources/Ignite/Modifiers/ModifiedHTML.swift
@@ -35,7 +35,7 @@ struct ModifiedHTML: HTML, InlineHTML, BlockHTML, RootHTML, NavigationItem {
             AttributeStore.default.merge(content.attributes, intoHTML: id)
         }
 
-        let modifiedContent: any HTML = modifier.body(content: self)
+        _ = modifier.body(content: content)
 
         AttributeStore.default.merge(content.attributes, intoHTML: id)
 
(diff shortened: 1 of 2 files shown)


erikbasargin:
> You're absolutely correct – thank you! This change conflicts with your other PR, but I have a whole batch of other adjustments waiting to go live. So, I'm going to close this issue, add your change locally, then push it up with all the other changes too.

I updated [the `fix-font-application` branch](https://github.com/erikbasargin/Ignite/tree/fix-font-application) with two new commits because my current fix stopped working starting from some latest changes.

- Return modified content from FontStyleModifier/body
- Merge attributes of unwrapped content when modified content is AnyHTML

twostraws:
@erikbasargin I've finished pushing up all the changes I was working on. I know @JPToroDev is currently looking at a few others, but right now fonts are definitely broken. I've posted a handful of other issues on the #Ignite channel on the HWS Slack.

twostraws:
@erikbasargin Sorry, I meant to add: if you have other changes you want to make, now would be a good time 🙂  If they are text-related, I'd tag @JPToroDev in case they clash with his own work.

PR #184, opened 2024-12-21
Fix bug where heading tags aren't properly applied to text.

JPToroDev:


JPToroDev:
When you merge this one in, you can squash it.

PR #198, opened 2024-12-25
Add optional spacing argument to `Section` init. Fix bug where `colum…

JPToroDev:
Add optional spacing argument to `Section` init. Fix bug where `column()` method has no effect.


twostraws:
I'm going to merge this in, but I definitely prefer the enum approach used in `Spacer` – we should move over to that at some point soon.

JPToroDev:
OK, that's an easy fix and a good point. Will circle back here shortly!

PR #201, opened 2024-12-27
Add `ResponsiveValue` enum and implement responsive overloads for `horizontalAlignment()` and `font()` modifiers.

JPToroDev:
Add `ResponsiveValue` enum and implement responsive overloads for `horizontalAlignment()` and `font()` modifiers.


twostraws:
This has some merge conflicts, and it also seems to bring in quite a few changes in one PR. I think once this and the other remaining one are merged, we *really* need to focus on writing tests so we can make sure our HTML generation is stable.

JPToroDev:
OK, I'll resolve the conflicts and break this into smaller commits where possible. After today, I'll shift my focus to testing.

JPToroDev:
I've move the `LengthUnit`-related code to a separate PR. Take a look at that first. Both should be good to go now.

PR #203, opened 2024-12-27
Add `defaultSiteWidth()` modifier...

JPToroDev:
Add `defaultSiteWidth()` modifier to complement the `screenRelativeFrame()` modifier, restoring content inside a full-screen `div` to the default site width.

Usage example:

```swift
Group {
    mySection
        .padding(.vertical, .px(100))
        .defaultSiteWidth()
}
.screenRelativeFrame(.horizontal)
.background(.blue))
```


twostraws:
I don't understand what this is trying to do – could you clarify?

JPToroDev:
I've updated the usage example with a missing modifier. The example code would create a section with a background that spans the entire width of the browser. But the content of the section would obey the theme's site width.

PR #208, opened 2025-01-03
Reintroduce background gradients.

JPToroDev:


twostraws:
I think this is another place where an `enum` might work better than several optionals. What do you think?

JPToroDev:
An enum instead of the static methods? Could you give me an example of what you mean?

PR #213, opened 2025-01-03
Decouple passthrough and container variants of `Group` into `Group` and `GroupBox`.

JPToroDev:
Like SwiftUI, `Group` is now a type that simply passes attributes to its children—how `Group(isTransparent:)` worked. `GroupBox` is a structural type that wraps its children in a `div`—how `Group()` worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ struct CustomContentLayout: ContentLayout {
         }
 
         if content.hasTags {
-            Group {
+            GroupBox {
                 Text("Tagged with: \(content.tags.joined(separator: ", "))")
 
                 Text("\(content.estimatedWordCount) words; \(content.estimatedReadingMinutes) minutes to read.")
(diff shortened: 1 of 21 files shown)


twostraws:
Could you perhaps outline the benefit of this change? Presumably if someone creates a `Group` and adds a class to it, that class propagates to all child elements?

JPToroDev:
> Presumably if someone creates a Group and adds a class to it, that class propagates to all child elements?

Yes exactly. If a modifier applies a class or style to an element behind the scenes, the `Group` propagates it to its children—that's it.

The way `Group` works now, wrapping its children in a `div`, can sometimes produce unexpected behavior. For example, if you use a `Group` inside a `Section` to set each child's `columnWidth` property, it currently will apply that width to the `div`, not each child, and the layout will be wrong.

Creating two distinct types—one for layout,another simply as a pass through—seems much less confusing and error prone than `Group` behaving in two completely different ways based on the value of `isTransparent`.

twostraws:
Yes, that makes sense. If we're going to proceed, could we perhaps think of a better name than `GroupBox`? I appreciate it already has a meaning in SwiftUI, but it doesn't quite match the HTML meaning, and I think also it's common enough to want to be shorter. I'm not keen on `Div` because it's just so meaningless – maybe just `Box`? Or `Block`? `Container`? Alternative suggestions most welcome!

JPToroDev:
I'm not sold on `GroupBox` either. And I have the same feelings as you about `Div`. `Container` was swimming in the back of my mind as an alternative—let's go with that!

JPToroDev:
All set.

PR #219, opened 2025-01-04
Update `ResponsiveValue` case names to mirror `LengthUnit` and save horizontal space.

JPToroDev:


twostraws:
? This seems like a step in the wrong direction – saving a few characters but losing meaning might be common on the web (vw, px, etc), but for places where we have a choice I think sticking with small, medium, etc would be clearer?

JPToroDev:
Good point. I'll remove that middle commit and simply make the full case names consistent with other small, medium, large, etc., phrasing elsewhere.

JPToroDev:
All set.

twostraws:
Thank you! 🙌

PR #224, opened 2025-01-07
Rename `Container` to `Stack`.

JPToroDev:
This PR renames the `Container` type to `Stack`. Here's the reasoning behind this change:

The original `Container` name had several drawbacks:
1. It could be confused with Bootstrap's `container` class, potentially leading to incorrect assumptions about its behavior—the main reason it wasn't my first choice initially.
2. The name didn't clearly communicate its purpose or functionality.
3. It introduced a standalone concept without established context in the Ignite ecosystem.

When I made my initial PR, `Container` seemed like the best option. But an unexpected moment of inspiration led me to `Stack` as a superior alternative because:
1. It builds upon our existing layout-related `Stack` types, providing better conceptual consistency.
2. It's more concise.
3. It avoids potential naming conflicts with Bootstrap conventions.

This rename improves clarity and better aligns with our existing type system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ struct CustomContentLayout: ContentLayout {
         }
 
         if content.hasTags {
-            GroupBox {
+            Stack {
                 Text("Tagged with: \(content.tags.joined(separator: ", "))")
 
                 Text("\(content.estimatedWordCount) words; \(content.estimatedReadingMinutes) minutes to read.")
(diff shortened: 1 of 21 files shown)


twostraws:
What is being "stacked" with a `<div>`?

twostraws:
Content in a `<div>` does not stack vertically. Could we perhaps give `Section` double duty?

twostraws:
Okay, then let's go for it. We really need to wrap up these big changes – I want to tag this thing and be stable!

PR #241, opened 2025-01-10
Fix `linkDecoration()` and improve clarity of implementation.

JPToroDev:


twostraws:
This has a merge conflict.

JPToroDev:
Resolved—thanks!

twostraws:
Maybe I'm misreading it, but does this remove the `linkDecoration()` modifier?

JPToroDev:
Yes, this replaces `linkDecoration(.heavy)` with `linkStyle(.underline(.heavy)).

This makes things clearer because 1) like text, links can have different decorations, but unlike `textDecoration()`, `linkDecoration()` only modifies the appearance of one decoration type, underline, and 2) users could use `linkStyle(.button)` and `linkDecoration(.heavy)` together, which doesn't make sense.

`linkDecoration()` was internal before, so this won't require any updating from the user's side.

Issue #57, opened 2024-05-20
Various Build Issues


nscalemodeller:
Running IgniteStarter from within Xcode doesn't appear to replace the Build folder after the first run. If I do a simple change to the 'Hello World' text on the Home page and rebuild the site the resulting index page doesn't reflect the change. I also get a “.DS_Store” couldn’t be copied to “Build” because an item with the same name already exists message.

I also downloaded and installed ignite as a CLI. It installed fine and creates a website but fails to build it. After a longish delay where the dependencies are downloaded and compiled I get a build failed message. Second run is faster but I get another error:

% ignite build
⚙️  Building your site...
warning: could not determine XCTest paths: terminated(1): /usr/bin/xcrun --sdk macosx --show-sdk-platform-path output:
    xcrun: error: unable to lookup item 'PlatformPath' from command line tools installation
    xcrun: error: unable to lookup item 'PlatformPath' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
    
[0/1] Planning build
Building for debugging...
[0/3] Write swift-version--1AB21518FC5DEDBE.txt
Build complete! (0.14s)

❌ Failed to build.

Is there something I am missing? Thanks!

nscalemodeller:
I managed to fix the CLI issue. I found this stack overflow post that fixed the issue:
https://stackoverflow.com/questions/61501298/xcrun-error-unable-to-find-utility-xctest-not-a-developer-tool-or-in-path/61725799#61725799

nscalemodeller:
An update on this.

If I run the package while it’s sitting within my iCloud folder I get this ( and another error ). However, if I drag the project to the desktop the project builds as normal and replaces the build folder correctly, as well as not producing the other error.

The extra error is that I have a simple markdown file sitting in the Resources folder that load fine while the project is located on the desktop but fails if I run the project from my iCloud folder. Strangely all other markdown files in the Contents folder load normally as do the Assets:

```
struct About: StaticPage {
    var title = “My Title”
    
    func body(context: PublishingContext) async -> [BlockElement] {
        let page = context.data(forResource: "about.md")!                    <—Thread 2: Fatal error: Unexpectedly found nil while unwrapping an Optional value
        let str = String(decoding: page, as: UTF8.self)
        
        Text(title)
            .font(.title3)
        
        Text(markdown: str)
    }
}
```

Document paths:

[...]

PR #246, opened 2025-01-12
Fix access to file paths: do not use percent encoding

pd95:
As shown in issue #57, space characters in a path can lead to build errors, due to percent encoding being used in URL paths. By using `.path(percentEncoded: false)` the original path can be retrieved.

Below is the code from #57 adapted to the new syntax

```swift
struct Home: StaticLayout {
    @Environment(\.decode) var decode
    var title = "Home"

    var body: some HTML {
        let page = decode.data(forResource: "about.md")!
        let str = String(decoding: page, as: UTF8.self)
        
        Text(title)
            .font(.title3)
        
        Text(markdown: str)
    }
}
```


twostraws:
Thank you!

Issue #64, opened 2024-05-26
build command not reporting error with layouts


jobearrr:
On my `Site` definition, I had the layouts assigned like below:
``` swift
var layouts = [
    BlogPost()
]
```
where `BlogPost` is a custom layout inheriting from `ContentPage`.

When I run the `ignite build` command, it finishes with the message `✅ Successfully built!`, but I look at the `Build/` folder and the contents using the `BlogPost` layout are not there.

I ran the executable scheme on Xcode and there I got an error `Failed to find layout named BlogPost.`

Should there be a difference between these two building options?
I think it would be nice to have the `build` command reporting that error which I got reported when running the executable scheme on Xcode.

jobearrr:
Yes, the problem was solved by specifying the type.

twostraws:
Thank you! I'm going to leave this issue open because I think we should document this on the Ignite Samples site, so this is a great first issue for someone.

jkaunert:
Xcode autocomplete adds the type when you accept, FYI.

twostraws:
Yes, but I don't think it's terribly clear to many people what the distinction is. I think we should definitely look to make this clear on IgniteSamples somewhere.

JPToroDev:
Addressed in IgniteSamples in the latest commit.

PR #249, opened 2025-01-13
Separate build and run commands to show runtime messages

pd95:
By separating the two phases we can show the runtime errors/warnings to the user.

Doing this finally shows the warnings of "IgniteSamples" also in the console:

```
Publish completed with warnings:
	- /images/photos/dishwasher.jpg: adding images without a description is not recommended. Provide a description or use Image(decorative:) to silence this warning.
```

Further it might have avoided issue https://github.com/twostraws/Ignite/issues/64 as the runtime warning "Your site must provide at least one layout in order to render Markdown." would have been shown to the user.


twostraws:
Great job!

PR #252, opened 2025-01-13
Fix inconsistent HTML output by changing collection type of various attributes from `Set` or `OrderedSet`.

jptorodev:


Sources/Ignite/Publishing/SiteConfiguration.swift
```diff
@@ -5,6 +5,8 @@
 // See LICENSE for license information.
 //
 
+import OrderedCollections
```
MrSkwiggs, line 8: This should not be necessary? (as per `CoreAttributes.swift:12` where the typealias was declared)

JPToroDev: Everything else comprehensible and working as expected?

MrSkwiggs:
I realise now it's not just about keeping the attributes in `OrderedSet`s; but it's also about making sure they're sorted correctly when rendering. I'll publish my own PR which already has all that in if you don't mind 🙏

JPToroDev:
Of course. This PR is just for reference. Will close it now. 🙌🏼

PR #253, opened 2025-01-13
Fix inconsistent use of content metadata identifier

MrSkwiggs:
This fixes an inconsistent use of the `modified`/`lastModified` content metadata identifier.

Also removes an unnecessary cast to String (using `.formatted(.iso8601)`) which creates issues when accessing `content.date` or `content.lastModified` which attempts to read a `Date`

### DIFF
--- a/Sources/Ignite/Framework/Content.swift
+++ b/Sources/Ignite/Framework/Content.swift
@@ -180,12 +180,12 @@ public struct Content {
         }
 
         if metadata["date"] == nil {
-            metadata["date"] = resourceValues.creationDate?.formatted(.iso8601) ?? Date.now
+            metadata["date"] = resourceValues.creationDate ?? Date.now
             hasAutomaticDate = true
         }
 
-        if metadata["lastModified"] == nil {
-            metadata["lastModified"] = resourceValues.contentModificationDate?.formatted(.iso8601) ?? Date.now
+        if metadata["modified"] == nil {
+            metadata["modified"] = resourceValues.contentModificationDate ?? Date.now
         }
     }
 
--- a/Sources/Ignite/Framework/Content.swift
+++ b/Sources/Ignite/Framework/Content.swift
@@ -43,7 +43,7 @@ public struct Content {
     /// The last modified date of this content. This might be the same as
     /// the publication date if the content has not subsequently been changed.
     public var lastModified: Date {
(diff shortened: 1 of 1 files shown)


MrSkwiggs:
@JPToroDev Could I get a set of 👀 on this? 🙏

twostraws:
Thank you! Dropping the `formatted()` call is excellent, but why are we also renaming `lastModified`?

MrSkwiggs:
@twostraws Because its use was inconsistent (one place we used `metadata["modified"]`, the other we used `metadata["lastModified"]`), which causes issues when accessing `content.lastModified` which itself tries to access `metadata["modified"]`.

```swift
/// The last modified date of this content. This might be the same as
/// the publication date if the content has not subsequently been changed.
public var lastModified: Date {
    metadata["modified"] as? Date ?? date
}
```

Unless I'm mistaken, `modified` & `lastModified` are supposed to be the same metadata values? 

I found out about this when my content pages would always show the same dates for `date` & `modified`, turned out we were setting the dates to `metadata["lastModified"] = contentModificationDate` but then trying to access it from `metadata["modified"]`

twostraws:
Ah! I see. Well, it's definitely wrong in the current implementation 😂   I didn't have the forethought to add any comments here (sorry!), so I'm just going to take a guess at what Past Paul meant here. I suspect my thinking was "some users might use 'modified' in their Markdown, some might use 'lastModified'; we should support both."

So, how does this sound as a potential solution?

- We look for `lastModified` as a parseable date string. If it exists, we put it in `metadata["lastModified"]` as a date.
- If that doesn't exist, we do the same for `modified`, again storing it in `metadata["lastModified"]` as a date.
- That means we support both options, but regardless put them both into `metadata["lastModified"]`, to be accessed through the `lastModified` property.
- The `fatalError()` ought to be placed with a thrown error.

MrSkwiggs:
Aaah ok haha 👍 Will do!

MrSkwiggs:
Actually I was thinking, do we _really_ want to support both? it might create confusion and make the documentation a bit fuzzy 🤔 

I'll replace the `fatalError()` in the meantime

MrSkwiggs:
Added support for both `modified` and `lastModified` in the end. Please have another look 🙏

PR #258, opened 2025-01-15
Fix environment values for content layouts

Mcrich23:
Hi All,
I noticed that the content layouts don't inherit the environment properly, which could make stuff like suggested articles more difficult. So I fixed it!


JPToroDev:
Excellent work—thank you!

twostraws:
Lovely!

PR #259, opened 2025-01-15
Simplify environment setup.

JPToroDev:
Thanks for the inspiration @Mcrich23 !

### DIFF
--- a/Sources/Ignite/Framework/Environment/EnvironmentReader.swift
+++ b/Sources/Ignite/Framework/Environment/EnvironmentReader.swift
@@ -8,7 +8,7 @@
 /// A protocol that allows types to read and write environment values.
 public protocol EnvironmentReader: Sendable {
     /// The current environment values for this reader.
-    @MainActor var environment: EnvironmentValues { get set }
+    @MainActor var environment: EnvironmentValues { get }
 
     /// The type of layout you want this page to use.
     associatedtype LayoutType: Layout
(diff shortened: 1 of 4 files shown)


twostraws:
Hmm. I really liked the `defer` use, because in theory it allowed us to nest environments in the future, similar to SwiftUI. By the looks of it that would be going away?

JPToroDev:
Could you elaborate a little more on what you mean by nested environments?

twostraws:
Anywhere when multiple values for the same setting go into the environment, meaning that different parts of the view hierarchy get different configurations.

```swift
Thing {
	Thing {
		Thing {
			Text("Whatever")
		}
		.font(.title1) // override the title just for this bit

		Thing {
			Text("Other")
		} // use the parent environment override here
	}
	.font(.title2) // etc
}
.font(.title3) // just for luck
```

JPToroDev:
Ah, I see. Your example is focused on the view-based environment, which is currently being handled via CSS inheritance within the modifiers themselves.

The site environment (aka `@Environment`) is really just storing global, static configuration values. Do you think there'd be a need to pass down other information in the view-based environment beside CSS styles like above?

twostraws:
SwiftUI's `@Environment` can read all sorts of things, including the currently active font. I guess it's a nice stretch goal!

JPToroDev:
I agree—a nice stretch goal and a fun challenge! 

I think our implementation of both environments would need to evolve quite a bit in that case. So whether we merge in or forgo this PR (either of which works for me), I think the code would change at that juncture.

PR #262, opened 2025-01-16
Fix compiler warnings.

JPToroDev:
The only way to avoid the new`@discardableResult` for `fontStyle()` would be to use the pattern:

```swift
var modified = content
modified = modified.fontStyle()
// etc.
```

Let me know if that is preferable.


twostraws:
Yeah, these warnings are annoying – thank you!

Issue #264, opened 2025-01-16
Support symbolic links in Content directory parsing 


wti:
This is to request/discuss/frame a PR for adding symbolic link support when parsing content.

**Motivation**:

The `Content` directory now requires users to put their markdown in the site/package directory (under its git).  

But users (and particularly teams) often have existing directories of content they'd like to publish, which we can support with symlinks.  This avoids Ignite getting tangled in the business of producing or assembling extrinsic markdown; it can just absorb content from upstream processes.

**Semantics**:

Currently symbolic links don't appear to be directories or markdown, so they are skipped.  

New behavior is to resolve symbolic links before checking attributes, but to consider the link name as the name for the resource.  The deployment path (i.e., the URL at which the resource should appear) needs to be built from these possibly-aliased names.

** Related **
Currently .MD or .markdown files also are missed.  Supporting symlinks would offer an (ugly) workaround of making symlinks with `.md`.  

However, it would be easy and better to add file suffixes as a site configuration option.

Symlinks do also enable users to manage the translation of filenames to urls, e.g., from case-sensitive filesystems (but not to exclude otherwise valid .md files).

**Compatibility and Deployment**
To ensure builds can reproduce prior behavior, it's possible to require a Site flag, defaulting to false, to enable the feature.   Similarly, for API users, new API's could use new parameters that default to existing behavior.

Testing is not easy.  It's hard to automate or even demo because symlinks are outside Swift's purview (though a shell script should work).  The varied platforms present some risk here: Swift's platform support for detecting symlinks is probably reliable on MacOS and Linux, but unclear on Windows.

Docs would require a short explanation of the behavior.

For implementation, see (upcoming) PR.

wti:
PR: https://github.com/twostraws/Ignite/pull/265

wti:
urk: this is only really valid for single links to directories

---

Geez, you'd think I'd check my assumptions!

It turns out: when resolving symbolic link, **all** symbolic links are traced, so a link to a link to a file/dir is indistinguishable from a link to the file/dir.  That means it's not possible to replicate the path the user would enter on the command line as the "deploy" path.

Thus, unless there's an API to resolve only the current leg of the symbolic link, this (at best) should be limited to single link to a directory, documenting the limitations that links, and the PR implementation is over-complicated.

wti:
There is an API: `FileManager.destinationOfSymbolicLink(atPath:)` will resolve each link step.  

[...]

PR #265, opened 2025-01-16
Content directory symlinks for #264; also permit some content errors, consider configurable suffixes

wti:
For https://github.com/twostraws/Ignite/issues/264

**Issue changes**
This reates a `ContentFinder` to gather content resources, resolving symlinks and maintaining a deployment path.

`Content` initializer now takes an optional deployPath, defaulting to nil for the prior behavior.

**Incidental changes**
This also adds some code features beyond the direct scope of #264:

`Content` also publishes the resource keys it expects.  That should simplify adding and removing value dependencies, since clients will get any updates.  Changing the parameter to support the published (static) keys as a default parameter should be possible.

This supports printing some content errors instead of failing on the first one.  That means a small content error can just print the error, instead of breaking the whole site build.  `Site` gets a new Int indicating how many errors to tolerate, defaulting to 0 for current behavior of throwing the first error.

`ContentFinder` adds (uppercase) ".MD" support and can be easily configured with more suffixes.  We could add a `suffixes: [String]` property to the SIte and make that user-configurable with the current default of ".md" only.

**Standards and Limitations**

This is already behind your recent commit (after I branched), but I thought it best to wait for feedback before final edits and rebasing.

I ran swift-lint and got 8 errors, but in other files.

Testing: This does not include a test shell script, and I haven't run it on the examples yet, only my own ~500+ files with multiple symlinks.

Apologies for the big ask and scope without prior discussion.  Symlinks can be tricky, and I definitely wouldn't want to create new stream of support requests, so "no" is a fine answer :)


wti:
Update: invalid implementation - see issue comment: https://github.com/twostraws/Ignite/issues/264#issuecomment-2599807160

Sources/Ignite/Publishing/PublishingContext-Finding.swift
```diff
@@ -0,0 +1,124 @@
...
+    ]
+
+    /// Find content input for a  content maker (and manage deploy path):
+    /// recurse from content root, resolve symbolic links, forward files matching any suffix.
+    ///
+    /// - For symbolic links, use the link name for constructing the deploy path and checking suffixes.
+    /// - The name used for the file deploy path is exclusive of the suffix matched.
+    ///
+    /// Limitations:
+    /// - Does not detect or avoid symbolic-link cycles
```
twostraws, line 31: Could this be resolved with some sort of Set of URLs that have been processed already? It seems a little dangerous as-is.

Sources/Ignite/Publishing/PublishingContext-Finding.swift
```diff
@@ -0,0 +1,124 @@
...
+    /// - Mixing errors for input, url/resource, and contentMaker
+    ///
+    /// - Parameters:
+    ///   - root: URL for readable directory
+    ///   - suffixes: non-empty Array of non-empty String; one must be the suffix of the filename
+    ///   - contentMaker: closure taking ``DeployContent`` and returning false to halt
+    func find(
+        root: URL,
+        suffixes: [String] = [".md", ".MD"],
```
twostraws, line 40: Catching uppercase and lowercase forms of this extension is very helpful, but I'm wondering whether a better approach might be to make the check itself case-insensitive? If users override this with (e.g.) .txt, I think they'd expect to match .TXT and other variants automatically.

Sources/Ignite/Framework/Content.swift
```diff
@@ -138,7 +141,9 @@ public struct Content: Sendable {
         title = parser.title.strippingTags()
         description = parser.description.strippingTags()
 
-        if let customPath = metadata["path"] as? String {
+        if let deployPath, !deployPath.isEmpty {
+            path = deployPath
+        } else if let customPath = metadata["path"] as? String {
```
twostraws, line 146: I don't fully understand the logic here: if users have set a `path` property in their content's metadata, I would have thought that should be honored regardless of what other logic we have. Why would `if let customPath = metadata["path"] as? String {` not come first here?

twostraws:
Thank you for this change! Separating content detection from the publishing context really helps clean up the code, and I absolutely agree that supporting symlinks will help a lot of people. I've added a handful of small comments/questions; if we can get those resolved, then I'm happy to merge this in.

Sources/Ignite/Publishing/PublishingContext-Finding.swift
wti: Yes.  

I'll save the directories visited (normalized using `standardizedFileURL`), in a Set as you suggest for faster lookup.

I think we can ignore files, even if they have been processed already.  I can imagine a user putting the same file at different locations via symlinks, so it's not clearly an error or a cycle to see it twice.  One might argue the same is possibly true for symlinked directories, but we'll just say no to that for simplicity now.

Conservatively this should probably be an error even though we can just prune the cycle.  It's likely not intended and indicates structural deploy issues.  If/since we will allow users to tolerate some errors, they have a workaround if it is intended.  But I'd also be happy to print warnings and continue.

That in turn raises the question whether the tolerance for content- and cycle-errors should be distinct settings.  It would be easy to add another Int parameter or wrap all error configuration in a struct/enum.  But we're already mixing error sources, and it would complicate things to have per-type tolerances, and the library aims to be simple, so I'm guessing no.

Sources/Ignite/Publishing/PublishingContext-Finding.swift
wti: Yes, we can match case-insensitively.

Also, I considered using a Set, but like the order and convenience of the array, which permits users to specify longer suffixes before shorter (`[".tmp.md", ".md"]`), to enjoy the benefit of longer clipping.  Let me know if you prefer swift-collection `OrderedSet`

(I haven't checked what happens when clipping or symlinks results in the same deploy path for two content URL's, but I assume that's something to manage properly.)

Sources/Ignite/Framework/Content.swift
wti: Sorry, yes: I will respect and document the priority of the more-specific document metadata.

Sources/Ignite/Publishing/PublishingContext-Finding.swift
```diff
@@ -0,0 +1,124 @@
+//
+// File.swift
```
MrSkwiggs, line 2:
```suggestion
// PublishingContext-Finding.swift
```

wti:
**PR updated** per feedback:
- metadata takes priority for deploy path
- suffix extraction is now case-insensitive.  (This is a behavior change.)
- Errors are thrown when a directory is visited a second time.  The error includes both paths to the directory.  This handles duplicates and cycles (both induced by symbolic links).  Errors are not thrown when a file is included twice by symbolic links.
- File comments all properly named.
- Error tolerance was not updated; it's just a counter defaulting to zero.

Also
- I removed overlapping URL-prefix derivation of deploy path in `Content.init(..)`

**Testing** took a fair percentage of effort, though it only targets `ContentFinder` (not any integration).  There are only 5 test cases, but we can create more by specification now as needed.  I used a fairly condensed style for test declarations and utilities, and tried to limit the scope (I'm a bit uncertain of the Ignite's overall testing direction).  It now tries to have the test driver and suite contain mainly high-level logic, with minutia in the helper, but that scatters the type members - hmm.  I'd grade it usable but not super clear or robust, and recommend vetting a bit more after review, when adding cases.

Did I backtrack on hygiene?
- swiftlint has ~40K warnings now; it was ~10 when I started ??
- I thought there was a swift-format configuration as well.

Deferred
- I realize we do no checking for uniqueness of the deploy path.  Users might appreciate a warning for duplicate paths (if it doesn't already throw an error).

wti:
Sorry, swift-lint warnings now fixed

wti:
Rebased on recent changes in main...

JPToroDev:
Thank you for staying persistent with this PR—I'm really excited about it! I'll be reviewing it thoroughly in the coming days and will circle back here with any questions. Based on your revisions though, it should be a straightforward merge.

Sources/Ignite/Publishing/PublishingContext-Finding.swift
JPToroDev: >  But I'd also be happy to print warnings and continue.

Could we go with this approach and eliminate the error tolerance setting?

Sources/Ignite/Publishing/PublishingContext-Finding.swift
JPToroDev: We use `OrderedSet` a lot throughout Ignite. Could we use it here too?

JPToroDev:
> I realize we do no checking for uniqueness of the deploy path. Users might appreciate a warning for duplicate paths (if it doesn't already throw an error)

Has that still been deferred or is that part of your recent updates?

JPToroDev:
I've added some comments throughout the PR. Overall the code looks good, but a lot of the naming isn't very reader friendly. Once those notes are addressed, I'll be very happy to merge this in—let me know if you have any questions.

Sources/Ignite/Publishing/PublishingContext-Finding.swift
wti: > We use OrderedSet a lot throughout Ignite. Could we use it here too?

I'm happy to, if you like, but I don't think it buys us anything, and it might confuse users who are just trying to list suffixes.  

All users understand arrays, and few users would specify duplicate elements in such a small, literal list (so they don't need a set).  Also, it's harmless if they do specify a duplicate since the code returns on first match.  So there's not much benefit to being a Set here.

`OrderSet`, by contrast, requires reading the documentation to see if it's using insertion order or the Comparable String order.  Since users rely on that to swap out longer variants of suffixes (`.tmp.md` above), that's key to understand.

Also `OrderedSet` doesn't solve what's likely to be the most common mistake, of needlessly supplying upper- and lower-case variants of the same suffix. (But again, that's harmless.)

wti:
>> I realize we do no checking for uniqueness of the deploy path. Users might appreciate a warning for duplicate paths (if it doesn't already throw an error)

> Has that still been deferred or is that part of your recent updates?

Yes, duplicate directories are now detected and result in an error listing both originating paths.

wti:
> but a lot of the naming isn't very reader friendly. 

Sorry, I'm happy to do renames as needed.  I tried to make public API super clear.  I also looked for unclear function-local names, and did find `resrcSet` which should be `resourceSet`.   I'll try to be clearer next time, but again if you have some blocking re-names, I'll do them!

> Once those notes are addressed, 

Please let me know if I missed anything.

JPToroDev:
Yes, the public API looks excellent. I'm very impressed with it. I'm really just talking about the test files, and it seems like they need extensive cleaning up. It's not even a stylistic thing—it's just taking a tremendous amount of effort to understand what's going on, and since this is a very significant addition, it's important that the tests are as crystal clear as the public API to the maintainers. Does that make sense?

wti:
> the test files [...] it's just taking a tremendous amount of effort to understand what's going on

Thanks for balking on that; I mentioned a ways back it was "not super clear or robust".  I could see the confusion better after forgetting them myself.  Sorry for not addressing it earlier.

For now I mainly added extensive comments, but also removed some duplication and started reporting missed errors.

It should be the case now that a maintainer can add a test case by copying an existing one and re-configuring the file-items (and, for the directory links, the expected deploy paths).  Another check is whether after one introduces an error into the code, (a) some test fails, and (b) the test feedback indicates what's failing clearly enough.

It's possible the static member declarations and even the FileItem enum could be replaced with simple String paths and a smart interpreter, but I experienced that as making it hard to re-use items and easy to make mistakes.  I expect if more test cases are necessary, they'd be easier to assemble with a small deltas on the current cases.

I hope you find this clearer, and sorry for the earlier confusion.

JPToroDev:
Thank you! I'll review the test code in the context of the comments in new couple days.

> I'm happy to skip and print warnings for bad symlinks since pruning invalid targets should result in a safely-built (if incorrect) Site.

> If you're okay going in this direction, let's do that and eliminate the maxErrorCount property from Site. If we find folks need something more robust, we can revisit it in the future, but I think simpler is better for the time being.

Was this included in the recent batch of changes?

wti:
>> I'm happy to skip and print warnings for bad symlinks...

>> let's do that and eliminate the maxErrorCount property from Site

> Was this included in the recent batch of changes?

No ... and sorry, but I'm confused by your question.

In my recent response to this question   (<https://github.com/twostraws/Ignite/pull/265#discussion_r1951687008>) I distinguished symlink/site errors which should be thrown from content errors which should be tolerated up to some limit.  (Default is 0, the current behavior)  

Your question implies the reverse: ignoring site errors and throwing on any content error.  

Again, the rationale with symlinks is that the site builder owns the symlinks which rarely change, but external authors own the content, and likely have no way to validate it, resulting in high frequency of invalid input.  Breaking a site build for content errors risks making Ignite appear to be unusable.

Also this text you quoted I don't recall and can't find in the PR comments or issue, so I wonder if I missed some agreement or decision somewhere:

> If you're okay going in this direction, let's do that and eliminate the maxErrorCount

In any case, though I stand by the recommendation and the current code, if you and Paul still want the reverse, I'll take out `maxContentErrors` (it's about 10 lines total?) and print the site errors (however, printing instead of throwing does make them untestable).

JPToroDev:
That's my bad. I made a handful of comments on GitHub mobile, and they failed to publish. They should be visible now, although most are likely outdated.

> I distinguished symlink/site errors which should be thrown from content errors which should be tolerated up to some limit. (Default is 0, the current behavior)

I like this very much. I take issue only with `maxErrorCount` because that number is completely arbitrary. Why are 3 errors an issue, but not 2? What is significant about 6 errors? 

If there's no meaning answer to those questions, it'd be much less confusing to either tolerate content errors and print them or not tolerate them at all.

wti:
>  I take issue only with maxErrorCount because that number is completely arbitrary. Why are 3 errors an issue, but not 2? What is significant about 6 errors?

Good question.  Right now, and after this is merged, we decided the number of `maxContentErrors` is 0.  But after merging, if the Site owner wants to change it to 1 or 10 based on their experience with content failures and people being upset about the Site build failing, they'll be able to.  Maybe there's a few known errors; maybe a bucketful of errors means something's wrong upstream, and the Site shouldn't build.  In any case, the site builder/maintainer/owner is the one with the information necessary to decide.

If this proves to be popular and site owners are complaining about not being able to distinguish content error types, or want to ignore Site structural errors, etc.  then that would factor into more error controls and a more general error-handling discussion.  If instead no one ever adjusts that value from 0, then yes, that's some code wasted and some configuration noise we can safely remove.

Having managed sites in the past and possibly the future (hence my motivation for the PR), I personally view it as essential, but I have nothing that says I'm a representative sample.

JPToroDev:
That all makes sense. I think the real solution lies in developing a more robust error handling system in general, but that will have to wait a bit.

That said, I'm still not sold on `maxErrorCount`'s current implementation. 

Would something like an `ignoreContentErrors` Boolean suit your needs? Or an enum that has different severity levels, with each case's raw value corresponding to an error limit?

As long as the implementation isn't built around arbitrary numbers, I'm in favor of it—although if you can wait, I think it'd be best to tackle this issue properly when we revamp Ignite's error reporting system.

JPToroDev:
Actually, we don't need to wait until we implement a new error handling system in Ignite to add `errorHandlingConfiguration`.

I'll make a PR with this `Site` property and make sure the type has a `maxContentErrors` property—although I think an enum would still be much more preferable if that suits your needs.

wti:
>  can make a PR adding this property to Site and make sure the type has a maxContentErrors property

It's in this PR already, no?  It's defined in the Site protocol with a default of 0, and used in PublishingContext when parsing content, and described in the  "incidental changes" paragraph of the initial PR message.

JPToroDev:
I zoomed out a bit too much. Let me refocus the conversation.

I take issue with `maxContentErrors` because it's based on arbitrary values and lacks baked-in direction for users of any skill level.

I'd much prefer an enum like `ContentErrorTolerance`. If that approach would suit your needs, could we please go with that approach?

Thank you!

wti:
I removed the (incidental) `maxContentErrors` from `Site` and content parsing.  (The only lint error is from main, which I take to be out of my scope). I'm hoping that's sufficient to merge this PR.

Error design belongs in another PR, but to follow up on your directions I also created and tested a `ContentErrorTolerance` enumeration.  It's harmless to include (modulo test noise), but you can also just omit the last commit when merging.  To incorporate it would require moving the enum into the main sources and integrating where the `maxContentErrors` property was before, in Site specification and content parsing.

The enum has predicates for deploy path, filename, error count, and error message (and composites thereof), anticipating some use cases:
- count threshold, to permit a few but balk on cascades
- filenames suffix, in case that encodes a different kind or source of content
- deploy path prefix, to permit errors in some symlink trees but not others
- error messages, to tolerate specific types
- combinations combine or negate the logic so users are not limited to one predicate

However, I tried to keep it simple and didn't address some logic:
- text matches are case-sensitive prefix, suffix, or contains checks only, not regular expressions or unified "contains"
- Need to decide to canonically render the error - get localizedDescription?
- composites permit weirdness: empty parts, negation of NONE, etc.

In terms of usability, our one caller has to maintain the URL, deploy path, count, and error in order to use these predicates; doable but not simple.  For `Site` users, direct predicates themselves are mostly understandable, but the composites are terse and steep (and still not complete).

In terms of integration and compatibility, the enum should be valid to mark Sendable.  Since users are mostly creating rather than handling the enum, adding more cases as the code evolves should not adversely affect source clients.

Tests are demonstrative but not complete.  

Overall I hope `ContentErrorToleranceTests.swift` gives you enough of a feel for the issues presented with such an enum to consider it in light of other error-handling needs.

JPToroDev:
Absolutely exceptional work with `ContentErrorTolerance`!

I think for the direction in which you've taken the enum, a more accurate name might be `ContentErrorPolicy`.

Rather than having cases for `anyOf` and `anyExcept`, what do you think about simply making the ultimate `Site` property:

```swift
 var contentErrorPolicies: [ContentErrorPolicy]
```

instead of 

```swift
var contentErrorPolicy: ContentErrorPolicy
```

> To incorporate it would require moving the enum into the main sources and integrating where the maxContentErrors property was before, in Site specification and content parsing.

Could you either do that in this PR, or open another PR that implements the `contentErrorPolicies` property?

> (The only lint error is from main, which I take to be out of my scope)

There are several spots throughout the code where you disable valid SwiftLint warnings with prompts like `// swiftlint:disable:previous function_body_length`. Instead of masking these warnings, you could address them?

Let me know if you have any questions—I'd love to get this merged in today. 

Again, fantastic job with this revision!

wti:
Glad you like it! Agreed on deferring content errors to another PR. 

Regarding `swiftlint:disable`, I left them because I think readers would prefer it, but there are alternatives.  It seemed too wordy to comment on each one, but I should have commented somewhere to the effect of...

In main Sources:

- `function_body_length` for `ContentFinder.find(..)`: has a comment `lint exception: using local closures to break up logic`.  Like `PublishingContext.generateThemeVariables(..)`, the function is internally decomposed with local functions.  Externalizing them and the closure context would be worse.

In Tests: I find clearer when the unshared names are local to the test or fixture types, and declarative case components indicate their contents:
- 6 `nesting`: namespacing types (particularly typealiases) avoids polluting the top-level namespace.  Most are typealiases (solvable instead with global `nesting` configuration `ignore_typealiases_and_associatedtypes`), but I also use 2 local Error types.
- `large_tuple`: this is a private field accessor of enum state that's only used nearby.  Breaking it up only multiplies switch access with no increase in clarity.  (The name is pretty condensed, but it's only used in 2 clear destructuring assignments.)
- `identifier_name` is disabled when constructing test cases where the name is the composite of the fields with `_` delimiters (solvable globally with `allowed_symbols`).  I believe this helps readability, and it makes it clearer how to assemble/construct new cases.

It could be helpful to add justifying comments to each exception, but that can get duplicative and I didn't see those in the other exceptions in the tree.  I did at least try to keep them narrowly scoped to one line or a few static declarations.  (FWIW, I think these are design issues with swift lint: they should distinguish public/internal/private API for rules like `nesting` and tuple size, and `function_body_length` should avoid including internal functions and multi-line constants.)

re: PR logistics, whether you remove the last commit for content errors or keep it in, the next PR can iterate on the errors in the Tests until the semantics are good, and then migrate them to the main code once the PR discussion settles.

JPToroDev:
All right, I've cleaned up the test files and prepped `Site` for you to reintroduce the content-error policy handling logic:

https://github.com/JPToroDev/Ignite/tree/021425b

If you can make a follow up PR today, then I'll merge the above branch in for you. Thanks!

PR #270, opened 2025-01-16
Rename `PlainText.swift` to `String.swift`, to better match the underlying element

gemmakbarlow:
Also adjusts some tests to contain an empty String (`""`) case.

Very minor updates, noticed while poking around the project.


JPToroDev:
Fantastic—thank you!

PR #271, opened 2025-01-16
Improve theme handling to prevent visual flicker and ensure immediate response to system color scheme changes.

JPToroDev:


twostraws:
Thank you! A couple of things:

1. Could the `themeSwitchingScript` code go into our resources as a file, rather than live in Swift code?
2. What is the SwiftLint results file doing in there? Also, you get different output from me, which led me to realize my SwiftLint is outdated 😅

JPToroDev:
2. Whoops! That shouldn't be there—will remove. (Glad it proved useful while it was around though! 😂)

1. Do you mean move the JS out of the HTMLHead file, but keep the output the same? Or relocate the JS in the output? I can't stand the JS being in the HTML output, but it's the unavoidable trick to making everything work smoothly.

twostraws:
JS: I mean to move that code into a separate file in the Ignite source code, similar to the global-rules.css file. Store all the text outside of Swift code, rather than making it a string. The end result – the built output – should remain the same.

JPToroDev:
Yeah, unfortunately not. When the JS is in its own file, it's not executed immediately, which causes the flickering. It's not quite as pretty, but this PR is the correct way to do it.

JPToroDev:
I didn't think so, but upon further investigation: if we move ignite-core.js from the body to the head and use the `defer` attribute with `<script>`, I think the behavior will be the same. I'll confirm, and update the PR if so. 🙌🏼

JPToroDev:
Sorry to report that won't work. The only reliable solution is keeping the script inline.

JPToroDev:
Great idea. Just made a commit with that change!

Sources/Ignite/Elements/HTMLHead.swift
```diff
@@ -105,5 +107,19 @@ public struct HTMLHead: RootHTML {
...
+            themeSwitchingScript
+        }
+    }
+
+    /// An inline script that handles theme changes immediately.
+    private static var themeSwitchingScript: Script? {
+        guard let sourceURL = Bundle.module.url(forResource: "Resources/js/theme-switching", withExtension: "js"),
+              let contents = try? String(contentsOf: sourceURL)
+        else {
+            return nil
```
twostraws, line 121: I don't think we should silently send back nothing if this script is missing when needed – if it's needed for the build then it should be required, and throw `PublishingError.missingSiteResource` if it's missing.

JPToroDev: Great idea—will add!

twostraws: Or 3: Create a new `addError()` method as a counterpart to `addWarning()`. We could handle these separately at the end of the build phase, perhaps with a ❌ emoji bullet point or something equally serious? Perhaps we could then use ⚠️ with warning bullet points. I see warnings as minor things the user screwed up, whereas this is something much bigger and will cause their site to behave badly.

JPToroDev: I like that. In that case, I'll create a type, something like `ErrorReporter` to replace the `warnings` of `PublishingContext` and properly encapsulate error management.

JPToroDev: Can you merge this in as is? I'll follow up immediately with the proper error handling, since that's starting to feel like it's own little PR.

PR #274, opened 2025-01-18
Boolean attributes & convenience html modifiers

MrSkwiggs:
## Description
This PR adds boolean attributes (similar to custom attributes, but without a associated value)
Included in the PR are the following convenience modifiers:
- `html.required()`: Adds the `required` boolean attribute
- `html.disabled()`: Adds the `disabled` boolean attribute

Use:
```swift
let element = HTMLBody().customAttribute(name: "my-boolean")
let output = element.render(...)

print(output) // prints: <body my-boolean></body>
```
```swift
let element = HTMLBody().customAttribute(name: "my-other-boolean", isEnabled: false)
let output = element.render(...)

print(output) // prints: <body></body>
```

## Tests
- Check that boolean attributes are sorted
- Check that disabling an attribute prevents it from being rendered in.


Sources/Ignite/Framework/ElementTypes/HTML.swift
```diff
@@ -143,8 +143,8 @@ public extension HTML {
 
     /// Adds a data attribute to the element.
     /// - Parameters:
-    ///   - name: The name of the data attribute
-    ///   - value: The value of the data attribute
+    ///     - name: The name of the data attribute
+    ///     - value: The value of the data attribute
```
MrSkwiggs, line 147: I added an extra tab-worth of indentation here so that parameters are correctly picked up by the documentation engine (they appeared under `Discussion` for me otherwise). Let me know if this needs to be excluded from the PR

JPToroDev:
Thank you for this!

It seems that boolean attributes are just custom attributes with a true or false value, right?

If so, could we simplify the approach a bit?

1. I think we can avoid create a new type `BooleanAttribute` by create an overload for `AttributeValue`.

```swift
    init(name: Property, isEnabled: Bool) {
        self.name = name.rawValue
        self.value = value
    }
```

2. And I think we can avoid a new method `booleanAttributes()` by creating an overload for 'customAttributes()` for `HTML`:

```swift
    func customAttribute(name: Property, isEnabled: Bool) -> Self {
```

 `CoreAttributes` and `HTML` already have a lot going on, so keeping things as streamlined as possible is a priority.
 
There are also a lot of methods in `HTML` that are public right now that should be private. I will update this ASAP.
 
For user-facing modifiers, we can't simply extend `HTML`. All public modifiers need to use `HTMLModifier` to work correctly with user-defined types, because it ensures attributes are processed correctly.

Let me know if you have any questions, or need any help revising!

MrSkwiggs:
> It seems that boolean attributes are just custom attributes with a `true` or `false` value, right?

Not exactly. They're attributes _without_ a value. 
`Input().booleanAttribute("required")` produces `<input required>` (not `<input required="true">`). 

The boolean part just dictates whether or not they appear in the rendered output.

`Input().booleanAttribute("required", isEnabled: false)` produces `<input>` without the attribute altogether.

> 2. And I think we can avoid a new method `booleanAttributes()` by creating a `customAttributes()` overload for `HTML`:
> 
> ```swift
>     func customAttribute(name: Property, isEnabled: Bool) -> Self {
> ```

That's a neat one, indeed I wasn't sure how to surface this api but this is much better 👌 
 
> For user-facing modifiers, we can't simply extend `HTML`. All public modifiers need to use `HTMLModifier` to work correctly with user-defined types, because it ensures attributes are processed correctly.

Funnily enough I did implement a modifier at first but then decided against it... will revert back :)

MrSkwiggs:
(for context, I decided to implement this after reading up on [Bootstrap Form Control](https://getbootstrap.com/docs/5.0/forms/form-control/#disabled))

JPToroDev:
> The boolean part just dictates whether or not they appear in the rendered output.

Thanks for the clarification! How about we create a new `enum` called `BooleanAttribute` similar to `Property`?

And then create an overload to `AttributeValue` that uses it.

```swift
    init(name: BooleanAttribute, isEnabled: Bool) {
        self.name = name.rawValue
        self.value = ""
    }
```

It seems like we should keep `booleanAttributes` in `CoreAttributes`.

And we can update the method I pitched before to this:

```swift
 func customAttribute(name: BooleanAttribute, isEnabled: Bool) -> Self {
 
 }
```
Which will create an `AttributeValue` and add it to our `booleanAttributes` property?

How's that sound?

MrSkwiggs:
Thanks for the feedback! Will make the edits.

For reference, in your comment before it was edited, you were asking if it made sense to include the `isEnabled` property as it essentially did nothing when set to `false`; 

I would argue it's still convenient to vend this so that other developers can create custom elements where this can be set dynamically (as it's much more annoying to add/not add a modifier instead)

```swift
var isEnabled: Bool

// Convenient
HTMLBody {}.customAttribute("required", isEnabled: isEnabled)

// Quite a bit less convenient
if isEnabled {
  HTMLBody {}.customAttribute("required")
} else {
  HTMLBody {}
}
```

MrSkwiggs:
Actually changing `AttributeValue.value` from `String` to `String?` makes it a bit annoying as every other place using this type now needs to check that the value is not nil. 
Screenshot 2025-01-19 at 10 46 30

IMHO, it doesn't make much sense to declare stuff like `aria-label` without a value 🤔 So either we throw an error in all those cases, or we keep the separate `BooleanAttribute` type. 

What do you prefer?

JPToroDev:
> I would argue it's still convenient to vend this so that other developers can create custom elements where this can be set dynamically (as it's much more annoying to add/not add a modifier instead)

Right, but Ignite is a build-time environment, not a run-time environment. I think it'd make little sense for someone to set an initial value for a boolean attribute on a build condition—and much more sense for them to check something on page load at run time and update the presence of the boolean attribute via JS, right?

JPToroDev:
> it a bit annoying as every other place using this type now needs to check that the value is not nil.

We can handle this with a computed property inside `Attribute` to make this easy.

> IMHO, it doesn't make much sense to declare stuff like aria-label without a value

We wouldn't allow folks to create an aria without a value. The methods for assigning name-value attributes like aria would have a required `value` property.

> or we keep the separate BooleanAttribute type.

Really examining `CoreAttributes` has revealed some type conflation here: CSS declarations and CSS attributes. I think the way to go is properly decoupling style declarations from attributes with a new `Declaration` type, and then giving `Attribute` the tools it needs to easily handle different values for `value`, including `nil.

Take a look at #282 . I think this should make what we discuss above very straightforward.

MrSkwiggs:
> Right, but Ignite is a build-time environment, not a run-time environment. I think it'd make little sense for someone to set an initial value for a boolean attribute on a build condition—and much more sense for them to check a condition on page load at run time and update the presence of the boolean attribute via JS, right?

I didn't mean to have this as a runtime dynamic thing. 

**Something more like this**

```swift
struct MyField: HTML {
  let name: String
  let isEnabled: String

  var body: some HTML {
    Text(title)
    TextField()
      .isEnabled(isEnabled)
  }
}

struct MyPage: HTML {
  let fields: [String: Bool] = [
    "name": false,
    "address": true,
    // etc.
  ]

  var body: some HTML {
    Section {
      ForEach(fields) { field in 
        MyField(name: field.key, isEnabled: field.value)
      }
    }
  }
}
```

It's not runtime, but is still relevant imo? 🤔

JPToroDev:
Your example doesn't use attributes in any form, which I think shows it's really not necessary. 

```swift
struct MyField: HTML {
  let name: String
  let disabled: Bool

  var body: some HTML {
    Text(title)
    TextField()
      .disabled(disabled)
  }
}

struct MyPage: HTML {
  let fields: (String: Bool) = [
    ("name", false),
    ("address", true)
    // etc.
  ]

  var body: some HTML {
    Section {
      ForEach(fields) { field in 
        MyField(name: field.0, disabled: field.1)
      }
    }
  }
}
```

Attributes are only handled in `disabled()`, which would be implemented something like this and scoped to `TextField`:

```swift
func disabled(_ disabled: Bool = true) → Self {
    if disabled {
        // previously `customAttribute()`, 
        self.attribute(.disabled)
        // would add `AttributeValue(_ name: BooleanAttribute) to `customAttributes`
        // where `.disabled` is a case of the `BooleanAttribute` enum
    }
}
```

So let's forgo any `isEnabled` argument or new type, as it adds unnecessary complexity, and stick with [the simpler approach above](https://github.com/twostraws/Ignite/pull/274#issuecomment-2600916095), building on #282. 🙌🏼

twostraws:
There's quite a lot of discussion here, but: `func disabled(_ disabled: Bool = true) → Self` would match what we already have for `required()`. We would also need something (maybe modifiers, maybe parameters in the initializer) for `selected` and `multiple` for select boxes, `checked` for checkboxes, `readonly` for text boxes, and `autofocus` for form fields.

twostraws:
What's the plan here, folks?

MrSkwiggs:
Sorry as mentioned things have been a bit hectic on my side the past few days. Now that #282 was merged I'll take a look and accommodate for changes 👍

MrSkwiggs:
Alright, changes are in and should better match the refactor introduced in #282. Please have a look again.

I've also introduced some convenience static lets under Attribute for common ones (such as `required`, `disabled`, `readonly` etc). Figured it's better than using Strings where typos can slip through. 

> There's quite a lot of discussion here, but: `func disabled(_ disabled: Bool = true) → Self` would match what we already have for `required()`. We would also need something (maybe modifiers, maybe parameters in the initializer) for `selected` and `multiple` for select boxes, `checked` for checkboxes, `readonly` for text boxes, and `autofocus` for form fields.

I've implemented `disabled`, `required` & `readonly` (where applicable) for Button & TextField. Does Ignite already have `Select`, `RadioGroup` etc? _I may have totally missed them 😵‍💫_

twostraws:
Thank you!

MrSkwiggs:
Sorry I realized I missed one thing in this, will quickly push a small PR

PR #276, opened 2025-01-18
Fix custom HTML elements not rendering their attributes

### LINKED ISSUE #275: Custom attributes on custom HTML elements do not render
I just noticed that my custom HTML elements strip out any custom attributes that are set on them.

## What I tried
```swift
struct MyElement: HTML {
  let title: String

  var body: some HTML {
    Tag("label") {
      title
    }
  }
}

let element = MyElement("Hello").customAttribute(name: "for", value: "test")
let output = element.render(context: ...)

print(output)
```

## What I expected
The element prints out with the custom attribute
```swift
/// <label for="test">Hello</label>
```

## What happened
The custom attribute is missing
```swift
/// <label>Hello</label>
```

## Other considerations
It might also happen with other types of attributes, although so far I saw that `class` does work as expected

MrSkwiggs:
Fixes #275

### DIFF
--- a/Tests/IgniteTesting/Elements/Attributes.swift
+++ b/Tests/IgniteTesting/Elements/Attributes.swift
@@ -109,4 +109,24 @@ struct AttributesTest {
             output == "<\(tag) data-bar=\"baz\" data-baz=\"qux\" data-foo=\"bar\" data-qux=\"foo\"></\(tag)>"
         )
     }
+
+    struct CustomElement: HTML {
+
+        let title: String
+
+        var body: some HTML {
+            Tag("label") {
+                title
+            }
+        }
+    }
+
+    @Test("Checks that custom HTML render custom attributes")
+    func test_custom_html_render_custom_attributes() async throws {
+        let element = CustomElement(title: "Hello")
+            .customAttribute(name: "for", value: "test")
+        let output = element.render(context: publishingContext)
+        
+        #expect(output == "<label for=\"test\">Hello</label>")
+    }
 }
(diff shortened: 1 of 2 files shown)


Tests/IgniteTesting/Elements/Attributes.swift
```diff
@@ -109,4 +109,24 @@ struct AttributesTest {
...
+    }
+
+    @Test("Checks that custom HTML render custom attributes")
+    func test_custom_html_render_custom_attributes() async throws {
+        let element = CustomElement(title: "Hello")
+            .customAttribute(name: "for", value: "test")
+        let output = element.render(context: publishingContext)
+        
+        #expect(output == "<label for=\"test\">Hello</label>")
+    }
```
MrSkwiggs, line 131: Not sure if this test should live here or in a file for custom html elements instead. Let me know what you prefer 🙏

JPToroDev:
Thanks for flagging this!

I see the issue. We need to create a proper modifier for `customAttributes` that uses the `HTMLModifier` protocol, which ensures we're always using a stable ID.

We should not change the default implementation for `render(context:)`.

Can you please revise using that approach?

JPToroDev:
Thanks for flagging this!

I see the issue. We need to create a proper modifier for `customAttributes` that uses the `HTMLModifier` protocol, which ensures we're always using a stable ID for custom types. #279 resolves this properly. 

We should not change the default implementation for `render(context:)`.

The test is something we could merge in, but right now, because the custom element has only a single primitive `HTML` element as the `body`, it's giving a false positive.

Let's circle back to this after the latest fixes are merged in and focus on implementing boolean attributes. In the meantime, I'm going to close this PR.

PR #279, opened 2025-01-18
Correct `HTML` extension access levels. Add public modifiers for adding styles, classes, events, aria, data attributes, and custom attributes. Prefix private variants with "_".

JPToroDev:


twostraws:
The addition of `EventType`/`onEvent()` is most welcome, but is mixed in with all the underscoring changes that I'm less comfortable with – it makes the code quite messy visually, and also feels like the kind of change that will cause confusion going forward.

JPToroDev:
Yeah, I'm not the biggest fan of the underscores either, but we do need a way to distinguish our private methods from our public methods for those like `class()`. 

We could make the private version `class(name:)` and the public `class()`, but that seems equally bad.

Part of me wonders if it's necessary to expose some of these methods at all now. Should we allow folks to add their own raw inline styles and classes in views, or is it better to let the framework manage everything at this point since it's matured quite a bit?

twostraws:
Perhaps we could wind back a little bit: what is the problem you're trying to solve here? What could can a user currently write that will cause problems?

twostraws:
> For example, people can do this:

Thank you; I can see the problem now. 

> What's more, there are a ton of overloads and helpers that shouldn't be exposed at all.

That sounds like a separate problem, and can be resolved as such.

I think we need to investigate this one further; the current solution is quite messy, and I'm not keen to proceed with it.

If the overloads and helpers changes can be pulled out to a separate PR, that would be helpful. The same goes for the `EventType/onEvent()` addition; if that could be pulled out too, we can merge those easy wins in no problem, then come back to the other problem once we have a slew of tests to help us keep things stable.

JPToroDev:
Sounds good, Paul. I submitted those PRs. These are the methods we'll have to address in the future and in the meantime, ensure no one applies to custom `HTML` elements while devising tests:

```swift
func `class`(_ newClasses: String?...) -> Self

func style(_ property: Property, _ value: String) -> Self

func id(_ string: String) -> Self

func aria(_ key: AriaType, _ value: String?) -> Self

func customAttribute(name: String, value: String) -> Self
```

PR #282, opened 2025-01-19
Decouple CSS attributes and CSS declarations.

JPToroDev:
Removing this conflation allows us to give each type distinct behaviors—initializers, `CustomStringConvertible` descriptions, etc.


MrSkwiggs (approved):

twostraws:
Switching to `.init(property: .width, value: "fit-content")` from `.init(name: "width", value: "fit-content")` is really nice, and renaming `AttributeValue` to `Attribute` is also fine.

Given the struct's name would now be `Attribute`, I wonder whether we could adjust the initializer to drop the `property` external parameter label, making `Attribute(.width, value: "fit-content")`?

Also, it feels fair to say that we should move across to the enum variant for all standard properties? So, `content.containerStyle(.init(property: "line-height", value: String(customHeight)))` would change in LineSpacing.swift, to use `.lineHeight`.

That being said – and I'm really trying to guess at your goal here, because I didn't fully understand it from your PR description – it seems you're trying to distinguish between HTML attributes and CSS styles.

So, `<a href="thisisanattribute" style="thisIsA: style;">whatever</a>` – is that correct? 

If so, I think the name `Attribute` continues to work, but `Declaration` is a bit unclear to me.

JPToroDev:
All set.

JPToroDev:
Also, a while back you expressed interest in renaming `CoreAttributes` since its scope had expanded, and I continue to be in agreement. 

Now feels like a good time to make that adjustment in a follow-up PR. Perhaps something like `CoreMarkup` or `Markup`?

twostraws:
Thank you! I'll merge this now.

Regarding `CoreAttributes`: I'm worried that we don't have time before Sunday. I've already merged some big changes from the last few days, and there are still more waiting for me to finish reviewing – I want to be sure everything is rock solid before bringing in lots of others.

JPToroDev:
OK, let's circle back here after the remaining PRs are merged in. 

I think #277 should be next on the list, because that will lay the groundwork to exposing some of these silent build failures, which I think is very important to get sorted out before the testing initiative.

twostraws:
I'm reading it now! Lots of changes, but mostly minor. I'll be through it soon 🙂

PR #283, opened 2025-01-19
Create `AriaType` to eliminate implicitly stringed type and improve safety.

JPToroDev:
…ety.


twostraws:
This is great – thank you!

PR #285, opened 2025-01-22
Improve clarity of `ResponsiveValue` API and improve parallelism with `LengthUnit`.

JPToroDev:


twostraws:
I read this through and wasn't sure about it, but I think on balance it's the right move. It does make the API a little wordier, but the result is an improvement.

JPToroDev:
Yes slightly wordier, but now it complements `LengthUnit` and looks distinct from the semantic `SpacingAmount`.

PR #287, opened 2025-01-22
Fix `Content` date parsing to support various formats.

JPToroDev:
Current implementation is causing an error in IgniteSamples.


twostraws:
Good catch! This would be a good test in our upcoming testing push.

PR #289, opened 2025-01-23
Expose all errors in publishing flow. Improve error handling.

JPToroDev:
I decided to use colored notebook emojis in the publishing summary for uniformity—inspired by this: https://stackoverflow.com/questions/40583721/print-to-console-log-with-color

We'll need to review which errors actually warrant a fatal error.


JPToroDev:
Missed a couple, will reopen shortly.

twostraws:
Yes to reviewing the fatal errors. I think ideally we catch almost all of them nicely as publishing errors so we can handle that in any way we want, but for now I'll merge this in.

PR #290, opened 2025-01-24
Make HTML helpers and overloads private.

JPToroDev:


twostraws:
Okay, so we're saying that using these handful of modifiers makes no sense on user-defined types, because they might contain multiple values inside them – is the style attached to them all? Just one thing? It's confusing, so by locking them away we're forcing them to apply them directly in their custom type. Correct? Should `class()` be included?

JPToroDev:
> is the style attached to them all? Just one thing? It's confusing, so by locking them away we're forcing them to apply them directly in their custom type. Correct?

Not quite. The methods we've made private are variants of these five core methods:

```swift
func `class`(_ newClasses: String?...) -> Self

func style(_ property: Property, _ value: String) -> Self

func id(_ string: String) -> Self

func aria(_ key: AriaType, _ value: String?) -> Self

func customAttribute(name: String, value: String) -> Self
```

Those five methods should give folks the tools they need to customize their HTML while 1) limiting confusion by keeping the API simple, and 2) limiting our overhead of negotiating private implementations (that forgo `HTMLModifier`) and public implementations (which require it).

We still need to bifurcate those five methods into public and private variants, because currently they have no effect on custom HTML elements.

twostraws:
As far as I can see, none of the five things you cited should have any effect on custom HTML elements. Heck, almost every modifier probably shouldn't, at least until we managed to environment modifiers implemented.

JPToroDev:
Most of the regular modifiers work as expected on custom HTML elements—those like `background()` and `frame()` for example. 

Those modifiers where you'd expect the modifier to pass its style to its children—like `font()`—already have an "environment" like behavior through CSS inheritance. So they work.

The reason I opened this PR is because folks were trying to use `customAttributes()` on their custom HTML element and found those attributes weren't present in the output, which is why I think it's important to address those five modifiers in particular.

JPToroDev:
`ModifiedHTML` wraps its custom HTML content in `div`, so that's where things are applied.

twostraws:
And that's not a possible solution for these other modifiers? Because that would make attaching `id()`, `class()`, etc, sensible.

JPToroDev:
It is possible. That's the big issue we're trying to solve:

Right now these five core modifiers are implemented _without_ the `HTMLModifier` protocol:

```swift
func `class`(_ newClasses: String?...) -> Self

func style(_ property: Property, _ value: String) -> Self

func id(_ string: String) -> Self

func aria(_ key: AriaType, _ value: String?) -> Self

func customAttribute(name: String, value: String) -> Self
```

Right now they are implemented like this:

```swift
func style(_ property: Property, _ value: String) -> Self {
    var attributes = attributes
    attributes.styles.append(.init(property, value: value))
    AttributeStore.default.merge(attributes, intoHTML: id)
    return self
}
```

In order to get them to work with custom HTML, we need to implement them _with_ the `HTMLModifier`.

So how about his, instead of worrying about retaining a "streamlined" variant of those modifiers that doesn't use the `HTMLModifier` protocol like above, we just update the existing modifiers to use `HTMLModifier`.

After revisiting `ModifiedHTML` earlier in the discussion, I can see that it's essentially just a pass-through entity for primitive types, so I don't think we need to worry about introducing overhead by using that implementation on primitive types.

twostraws:
Yes, that's exactly what I'm thinking: move them all across to use `HTMLModifier`. The only hiccup here is that we want to make sure that simple elements don't get wrapped in an extra `<div>`, because that would cause all sorts of problems.

JPToroDev:
They won't :)

```swift
func render() -> String {
        if content.isPrimitive {
            AttributeStore.default.merge(attributes, intoHTML: content.id)
            return content.render()
        } else {
            let rawContent = content.render()
            var attrs = attributes
            if attrs.tag == nil { attrs.tag = "div" }
            return attrs.description(wrapping: rawContent)
        }
    }
```

Will get on the revision!

JPToroDev:
That issue will be separate from this PR though, so I'll resolve these conflict so this one can be merged in.

JPToroDev:
So there is one variable I forgot about until going back into the weeds, relating to name collisions.

If we simply convert the existing modifiers to use `HTMLModifier`, we won't be able to have fine grained control like this:

```swift
// private
func `class`(_ newClasses: String?...) -> Self {
// public
 func `class`(_ newClass: String) -> some HTML {

// private 
 @discardableResult func data(_ name: String, _ value: String?) -> Self

// public 
func data(_ name: String, _ value: String) -> some HTML
```

I mention this because it really doesn't make sense to make `value` or `newClasses` `String?` for the public or make a result discardable, and including either of those pairs of methods confuses the compiler.

Is this a minor annoyance we can look past or something that requires a better solution?

twostraws:
It's much better than introducing the underscores 😅

JPToroDev:
And also preferable to public variants like `class(name:)`, yeah?

twostraws:
Yes. If folks really need the `Self` type coming back – e.g. for `Image` – then they should rearrange their modifiers, like they'd do with SwiftUI.

JPToroDev:
Great—will have the PR done shortly!

JPToroDev:
I'm sorry to keep returning here, but it seems that this approach is not quite as straightforward or advisable as it seems on the surface. It goes much deeper than just stylistic considerations.

We have a few of elements that conform to both `BlockHTML` and `InlineHTML`, which is confusing the crap out of the compiler, and different fixes introduce different problems, creating a game of whack-a-mole.

What's more, things like this aren't possible anymore:

```swift
 private func renderSimpleItem(_ item: any InlineHTML) -> Section {
        Section(item)
            .class(getColumnClass(for: item, totalColumns: columnCount))
    }
```

We'd end up reworking a big chunk of the code base, limiting our flexibility behind the scenes, and introducing workarounds that aren't better than underscores—perhaps even worse.

If you want, I continue hacking through this, but I think the cost-benefit scale is starting to tip.

JPToroDev:
OK, I think I've worked out a solution. If we make those five methods part of the `HTML` protocol proper like this:

```swift
    associatedtype Output = Self
    func id(_ id: String) -> Output
```
And make our default implementations return `some HTML` so that they can be used with custom HTML, we can then add implementations like this to our primitive types:

```swift
 public typealias Output = Self
    public func id(_ id: String) -> Output { // we can extract the body into a method on `CoreAttributes` for reusability 
        var attributes = attributes
        attributes.id = id
        AttributeStore.default.merge(attributes, intoHTML: id)
        return self
    }
```

If we're gonna be adding code anyway, I think this let's us have our cake and eat it too.

PR #293, opened 2025-01-24
Add more test and fix missing implementation

MrSkwiggs:
Sorry I just realised I forgot to add `func disabled(_:)` to Button after wondering why my tests were not compiling


twostraws:
Thank you!

PR #296, opened 2025-01-24
Clean up attributes.

JPToroDev:
A few review items I couldn't address before the merge.


Sources/Ignite/Framework/BooleanAttribute.swift
```diff
@@ -0,0 +1,93 @@
...
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+/// An enumeration of HTML boolean attributes.
+///
+/// Boolean attributes are attributes that can be present or absent on an HTML element,
+/// where their presence indicates true and their absence indicates false.
+enum BooleanAttribute: String {
```
twostraws, line 12: Has this file moved?

JPToroDev: Yes, that file was accidentally included in the original PR with unused code.

twostraws: I might just be reading GitHub wrongly, but it seems that the file has moved up one directory, which is a strange choice?

JPToroDev: It was originally in the Rendering folder, now it's in the Framework folder.

Sources/Ignite/Rendering/Attribute.swift
```diff
@@ -30,14 +35,3 @@ public struct Attribute: Hashable, Equatable, Sendable, Comparable, CustomString
...
-// MARK: - Standard attributes
-
-public extension Attribute {
-    static let disabled: Attribute = .init(name: "disabled")
-    static let required: Attribute = .init(name: "required")
-    static let readOnly: Attribute = .init(name: "readonly")
-    static let hidden: Attribute = .init(name: "hidden")
-    static let checked: Attribute = .init(name: "checked")
-    static let selected: Attribute = .init(name: "selected")
-}
```
MrSkwiggs, line 43: I'm not convinced removing this and reintroducing the new `BooleanAttribute` makes sense here;
In general, I'd tend to avoid public-facing enums as any case addition/removal will cause compilation errors for our consumers...

Plus, we could have kept adding convenience constructors for other standard, non-boolean attributes here as well, such as 
```swift 
static func type(_ type: String) -> Attribute {
  Attribute(name: "type", value: type)
}
```

MrSkwiggs: I like the static func approach as well because it allows us to be strongly typed for specific attributes where only a set amount of values make sense. i.e

```swift
extension Attribute {
  struct `Type` {
    let name: String

    static let button: Self = .init(name: "button")
  }

  static func type(_ type: Attribute.`Type`) -> Attribute {...}
}
```
If that makes sense?

MrSkwiggs: > Making `BooleanAttribute` into a struct means we’d need to add a new `booleanAttributes` property to `CoreAttributes`, which means we’d need to update the `description()` method, etc. 

That's not what I meant. 
We already have the `Attribute` struct, let's keep using it and have its value be optional like you had previously mentioned 👍 

In fact I'm suggesting to remove the newly introduced `BooleanAttribute` enum and only keep `Attribute`.

There would be no need to change anything really, no refactor whatsoever.

> We’re adding complexity to something that already has a lot going on for no real discernible benefit. What's more, aria attributes and data attributes don't have their own structs, which would make things inconsistent. Keeping things simple is a consummate and worth priority.

I'm really only suggesting to keep the existing `Attribute` struct, and keep these convenience static constants that reduce chances for typos when repeatedly adding common attributes such as `type`, `disabled`, `readonly` etc.

> I don’t follow your example. Could you explain what you mean by this and the problem it solves? It sounds like some of these might be interesting hypotheticals, but I'm failing to see the practical use cases here. Attributes really aren’t that big a part of the public-facing framework, and honestly, I don’t think that’s something we want to change or encourage.

The same way we have convenience static constants for "boolean" attributes i.e `disabled`, we can also have static funcs for attributes with a value, as demonstrated in my example with `type=<value>`. For the type attribute, there are really only a few values that are generally used, i.e `button`, so instead of having it typed out in a String, we can vend variables instead. Define it once, and have the benefit of having a compile-time check done

JPToroDev: Thank you for the thorough explanation—it's put us back on the same wavelength. What you've outlined above simplifies things—so it goes without saying I love it!

Let's stick with the static property approach and ditch `BooleanAttribute`. My latest commit addresses this—it also handles `hidden` via a method, since the attribute recently graduated from boolean to enumerated.

twostraws:
If you fix the merge conflict, I'll merge this in now. However, this needs to be the final change before the testing event, because I want time to check Ignite Samples thoroughly before everyone joins.

JPToroDev:
All set.

JPToroDev:
Yes agreed—all set.

PR #297, opened 2025-01-24
Fix Access Control Using `HTML` Protocol Requirements

JPToroDev:
A partial example of the solution I mentioned in my comments. I think either this or underscores is honestly our best approach. Doing one set of modifiers using the `HTMLModifier` implementation is a slippery and likely dangerous slope.


JPToroDev:
All right, I've pushed this to its complete form, and I think it's the way to go. We can avoid any unpleasant API syntax like underscores, all while allowing different implementations for custom types and primitive types. If everything looks good with you, it should be ready to merge as is. 

Phew!

PR #300, opened 2025-01-26
Added tests for an extension on URL

Aradroid:
Just some tests for the `URL.removingWWW` method.


twostraws:
Please pull the latest version of the repository – you'll see there's a clear file structure now, where these tests will need to go.

PR #316, opened 2025-01-26
[Tests] Added basic Alert tests

OliverForth:
Alerts with different roles:
- primary
- secondary
- success
- danger
- warning
- info
- light
- dark


twostraws:
Perhaps this could be parameterized in the future? Worth evaluating.

PR #318, opened 2025-01-26
Mediaquery tests

MrSkwiggs:
Adds a bunch of MediaQuery tests

### DIFF
--- a/Tests/IgniteTesting/Modifiers/MediaQuery.swift
+++ b/Tests/IgniteTesting/Modifiers/MediaQuery.swift
@@ -107,4 +107,17 @@ struct MediaQueryTests {
             output == testCase.output
         )
     }
+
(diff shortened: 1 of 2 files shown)


Tests/IgniteTesting/Modifiers/MediaQuery.swift
```diff
@@ -0,0 +1,123 @@
...
+        )
+    }
+
+    @Test("Test contrast queries", arguments: [
+        (query: MediaQuery.contrast(.high),
+         output: "prefers-contrast: more"),
+        (query: MediaQuery.contrast(.low),
+         output: "prefers-contrast: less"),
+        (query: MediaQuery.contrast(.reduced),
+         output: "prefers-contrast: less")
```
MrSkwiggs, line 62: Not sure these should produce the same output?

PR #321, opened 2025-01-26
Tests for MarkdownRenderer

aerickson14:
Add some initial tests for markdown renderer

### DIFF
--- a/Tests/IgniteTesting/Framework/MarkdownRenderer.swift
+++ b/Tests/IgniteTesting/Framework/MarkdownRenderer.swift
@@ -15,13 +15,12 @@ import Testing
 @MainActor
 struct MarkdownRendererTests {
 
-    @Test("Markdown headers from string", arguments: ["# Header 1", "## Header 2", "### Header 3"])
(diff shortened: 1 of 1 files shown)


Tests/IgniteTesting/Framework/MarkdownRenderer.swift
```diff
@@ -14,8 +14,42 @@ import Testing
...
+        let element = MarkdownToHTML(markdown: markdown, removeTitleFromBody: false)
+
+        let content = markdown.replacingOccurrences(of: "> ", with: "")
+        #expect(element.body == "<blockquote><p>\(content)</p></blockquote>")
+    }
+}
+
+extension MarkdownRendererTests {
+    private func numberOfHashtags(in markdown: String) -> Int {
+        return markdown.filter { $0 == "#" }.count
```
aerickson14, line 53: This would fail for a test like "# Header with a #hashtag". It might be worth me updating this helper function to only look for `#` at the beginning of the markdown

aerickson14: This should now be resolved in the latest commit

PR #323, opened 2025-01-26
DONE: added basic Include test

AlejandroLanza:
Basic include test added, checks for output == ""


twostraws:
We talked about checking the build warnings were correctly introduced, but for now I'm going to merge this in – thank you!

PR #325, opened 2025-01-26
Add Card.Style testing

pd95:
A non-parameterized test as `Card.Style` is not sendable.


pd95:
Finally fixed the random errors :-)

PR #328, opened 2025-01-26
Added Tests for `Section`

Diggory:
Added a couple of tests for the initialisers for the `Section` element.


Diggory:
The initialiser for Section `init(_ items: any HTML, background: Color? = nil)` doesn't appear to use the colour...

PR #341, opened 2025-01-26
Tests for TextField

Rimesh:
Added test for TextField
1. TextField With Placeholder
2. TextField with Placeholder and Label

Test to check correct TextField Text Type
1. TextField with standard text input
2. TextField with email address input
3. TextField with password input
4. TextField with phone input
5. TextField with url input
6. TextField with search input
7. TextField with number input

Tests to modify the TextField
1. TextField is required
2. TextField is disabled
3. TextField is read only

### DIFF
--- a/Tests/IgniteTesting/Elements/TextField.swift
+++ b/Tests/IgniteTesting/Elements/TextField.swift
@@ -34,73 +34,13 @@ struct TextFieldTests {
         """)
     }
 
-    @Test("TextField with standard text input")
-    func textFieldWithStandardTextInput() async throws {
-        let element = TextField("Paul", placeholder: "Enter your name here")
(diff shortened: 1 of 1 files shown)


Tests/IgniteTesting/Elements/TextField.swift
```diff
@@ -14,8 +14,123 @@ import Testing
...
+        """)
+    }
+
+    @Test("TextField with Placeholder and Label")
+    func textFieldWithPlaceholderAndLabel() async throws {
+        let element = TextField("Paul", placeholder: "Enter your name here")
+        let output = element.render()
+
+        #expect(output == """
+        <input placeholder="Enter your name here" type="text" class="form-control" />
```
Rimesh, line 33: Produced HTML does not care about the Label passed.

PR #350, opened 2025-01-26
Eliminate final linting errors.

JPToroDev:


twostraws:
I'm taking this PR title to mean there will never be any more SwiftLint problems. Good luck 🙂

PR #359, opened 2025-01-26
Test for Font Style

NigelGee:
Looking at the `enum Style` for Font there is a lead for larger variant of body text using Bootstrap's large font size (1.25rem), so I excluded this one.


twostraws:
Thank you!

PR #362, opened 2025-01-26
Test for Font Weight

NigelGee:


Tests/IgniteTesting/Modifiers/FontWeightModifier.swift
```diff
@@ -14,8 +14,18 @@ import Testing
 @Suite("FontWeightModifier Tests")
 @MainActor
 struct FontWeightModifierTests {
-    @Test("ExampleTest")
-    func example() async throws {
+    @Test("Font Weight Test")
+    func fontWeight() async throws {
+        for font in Font.Weight.allCases {
```
twostraws, line 19: Would this work as a parameterized test?

NigelGee:
Did you mean like this?

twostraws:
That's an improvement, but I think we can read the raw values directly from the font weight rather than passing them into the test.

NigelGee:
This is better. Sorry first time doing this

Tests/IgniteTesting/Modifiers/FontWeightModifier.swift
```diff
@@ -14,8 +14,15 @@ import Testing
 @Suite("FontWeightModifier Tests")
 @MainActor
 struct FontWeightModifierTests {
-    @Test("ExampleTest")
-    func example() async throws {
+    @Test("Font Weight Test", arguments: Font.Weight.allCases)
```
twostraws, line 17: This is now perfect.

Tests/IgniteTesting/Modifiers/FontStyleModifier.swift
```diff
@@ -16,15 +16,13 @@ import Testing
 struct FontStyleModifierTests {
     @Test("Font Style Test")
     func fontStyleExcludingLead() async throws {
-        for font in Font.Style.allCases {
```
twostraws, line 19: FontStyle tests shouldn't be in this PR; this might have been added by accident.

NigelGee: No it should have been different but seem to add it to current open PR. Not sure how to back it out.

twostraws: The easiest thing to do is close this PR, update your fork to the latest version, and reapply your change. Make sure you take a copy of the test you actually want to include.

PR #364, opened 2025-01-26
More MarkdownRenderer tests

aerickson14:
Added a few more tests for `MarkdownRenderer`:
- metadataIsProcessed
- removeTitleFromBody
- convertTableToHTML

There was also a minor update to rename header to heading in these tests


twostraws:
Thank you!

PR #365, opened 2025-01-26
Add tests for EventModifier

vdeep:
Here I'm just testing the output for the `.onEvent` modifier. I'm re-using the output of the `EventType` and `Action` to compare the modifier output instead of comparing with strings.

### DIFF
--- a/Tests/IgniteTesting/Actions/EventModifier.swift
+++ b/Tests/IgniteTesting/Actions/EventModifier.swift
@@ -32,6 +32,10 @@ struct EventModifierTests {
         [ShowModal(id: "foo"), HideElement("qux")]
     ]
 
+    init() throws {
+        try PublishingContext.initialize(for: TestSite(), from: #filePath)
+    }
+
     @Test("Check if events are added correctly", arguments: tags, await Array(zip(events, actions)))
     func onEventAddsEventsCorrectly(tag: String, eventActions: (EventType, [any Action])) async throws {
         let element = Tag(tag) {}
--- a/Tests/IgniteTesting/Actions/EventModifier.swift
+++ b/Tests/IgniteTesting/Actions/EventModifier.swift
@@ -24,7 +24,7 @@ struct EventModifierTests {
         EventType.mouseOver
     ]
 
-    private static let actions: [[any Action]] = [
+    private static let actions: [[Action]] = [
         [ShowAlert(message: "foo")],
         [ShowAlert(message: "qux"), DismissModal(id: "baz")],
         [CustomAction("document.writeline('bar')")],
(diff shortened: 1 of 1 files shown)


twostraws:
This is a great start – thank you! Could this be a parameterized test?

vdeep:
I tried to do parameterized tests, but it required Sendable conformance for `EventType`. Would really appreciate your feedback on how this can be improved.

twostraws:
Try now 🙂

vdeep:
Updated the test to use `arguments`.

vdeep:
Getting the same error after adding that code block, but intermittently

image

vdeep:
Also got this failure when continuously running the tests
Screenshot 2025-01-26 at 7 50 58 PM

Could this be because we are only using first-four letters of hash for the Tag's id? I tried changing this to 8 characters and didn't face that failure again.
image

JPToroDev:
> Could this be because we are only using four letters from the hash for the Tag's id?

`truncatedHash` relates to an Ignite id used internally behind the scenes, not an HTML id. Could you try setting `.id("myID") on your `Tag`, which will set the HTML id to something consistent?

vdeep:
Same behavior after adding `.id("testID")`.

image

---

> truncatedHash relates to an Ignite id used internally behind the scenes, not an HTML id. Could you try setting .id("myID") on your Tag, which will set the HTML id to something consistent?

Actually I was talking about the Ignite internal ID. Please correct me if I'm wrong, but attribute array is being added to a singleton based on the tag's internal ID, right?

image

JPToroDev:
Yep, that's exactly correct. Perhaps there is a uniqueness issue with just four digits—but that should yield around 10,000 unique identifiers, so I'm skeptical. Rather than a screenshot, could you share the specific line(s) of code causing the failure? And are you still getting `nil` from the publishing context?

vdeep:
I'm not getting `nil` from the publishing context anymore.

For the above error, I don't normally get it. I get it only when I run the `EventModifierTests` repeatedly until failure with Relaunch option. The line I get error on is the `#expect` assertion (line `49`).

JPToroDev:
> I'm not getting nil from the publishing context anymore.

Well that's good!

When it fails, what's the specific discrepancy between the output and the expected string? The ID?

vdeep:
The failure message shows that the attributes are getting mixed up.

For example, I'm testing events with actions as follows:

```swift
private static let events: [EventType] = [
    EventType.click,
    EventType.doubleClick,
    EventType.focus,
    EventType.load,
    EventType.mouseOver
]

private static let actions: [[Action]] = [
    [ShowAlert(message: "foo")],
    [ShowAlert(message: "qux"), DismissModal(id: "baz")],
    [CustomAction("document.writeline('bar')")],
    [CustomAction("document.writeline('foo')"), ShowElement("qux")],
    [ShowModal(id: "foo"), HideElement("qux")]
]
```

These are `zip`ped and then passed as arguments to the test.

The error message shows that the tag contains different actions than what it was supposed to. For example, the tag with `doubleClick` event should only have the `ShowAlert` and `DismissModal` events, but it also contains another event, like the `load` event and its actions - `CustomAction` and `ShowElement`.

Here is the test output:

```
􀢄 Test "Check if events are added correctly" recorded an issue with 2 arguments tag → "body", eventActions → (Ignite.EventType.doubleClick, [Ignite.ShowAlert(message: "qux"), Ignite.DismissModal(id: "baz")]) at EventModifier.swift:49:9: Expectation failed: (output → "<body ondblclick="alert('qux'); const modal = document.getElementById('baz');
const modalInstance = bootstrap.Modal.getInstance(modal);
if (modalInstance) { modalInstance.hide(); }" onload="document.writeline(\'foo\'); document.getElementById('qux').classList.remove('d-none')"></body>") == ("<\(tag) \(eventOutput)=\"\(actionOutput)\"></\(tag)>" → "<body ondblclick="alert('qux'); const modal = document.getElementById('baz');
const modalInstance = bootstrap.Modal.getInstance(modal);
if (modalInstance) { modalInstance.hide(); }"></body>")
```

New tag is being created for each event, so one tag should not contain more than one event.

JPToroDev:
Perfect, thank you for the detailed information. I'll take a look the next time I'm at my computer and help you iron this out. 🙌🏼

vdeep:
Thanks @JPToroDev . 

By the way, this issue is reproducible in `AttributesTest` also. Try running the tests repeatedly until failure, with Relaunch option enabled.

JPToroDev:
Will do—thanks!

twostraws:
Thank you!

JPToroDev:
I think you've made a very astute observation pinpointing this issue to `HTML`'s `id` and `truncatedHash`. 

It seems that extra attributes are sometimes appearing because `truncatedHash` is not generating an `id` that is unique enough, and two elements from a test are being registered in `AttributeStore` with the same `id`. 

I've update `truncatedHash` in #413 to use 5 digits instead of 4, and the test failures have disappeared—like you observed using 8 digits.

Truly excellent work—thank you!

vdeep:
Thanks @JPToroDev !

PR #367, opened 2025-01-27
added tests for Date-ISO8601 and Date-RFC822

jaywardell:
I used Date.asISO8601 to create a list of 100 ISO8601 date strings.
I then added a test that loads that list, creates a Date from it, and then tests that date's .asISO8601 against the original String.

I then did the same process for Date.asRFC822


Tests/IgniteTesting/Extensions/String-KebabCased.swift
```diff
@@ -14,8 +14,86 @@ import Testing
 @Suite("String-KebabCased Tests")
 @MainActor
 struct StringKebabCasedTests {
-    @Test("ExampleTest")
-    func example() async throws {
 
+    /// Some types of string will have an output that's the same as their input.
+    /// Test examples of each of those cases
+    @Test("Noop Cases", arguments: [
```
twostraws, line 20: Should these parts be in here?

Tests/IgniteTesting/Extensions/Date-ISO8601.swift
```diff
@@ -14,8 +14,118 @@ import Testing
...
+3861-03-09T00:29:10Z
+0863-05-11T06:00:31Z
+3426-04-07T15:38:37Z
+"""
+        .components(separatedBy: "\n")
+
+    @Test("Test Against 100 ISO8601 Strings")
+    func asISO8601_outputs_proper_string() async throws {
+
+        let formatter = ISO8601DateFormatter()
```
twostraws, line 124: Rather than writing code to convert this on the fly, I would prefer to trim the list of dates back to just 10 or so, but then hard-code the correct response for them. I'd rather adding other possible points of failure here, so let's just check input string matches output string.

Tests/IgniteTesting/Extensions/Date-RFC822.swift
```diff
@@ -14,8 +14,121 @@ import Testing
...
+Fri, 04 Apr 3034 13:06:07 -0400
+Fri, 31 Oct 1597 06:38:21 -045602
+"""
+        .components(separatedBy: "\n")
+
+    @Test("Test Against 100 RFC822 Strings")
+    func asRFC822_outputs_proper_string() async throws {
+
+        let formatter = DateFormatter()
+        formatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss Z"
```
twostraws, line 126: Again, please trim the input back to 10 varying examples, then hard-code the response rather than calculating them in the test. Thank you!

Tests/IgniteTesting/Extensions/String-KebabCased.swift
jaywardell: oops. sorry. will resubmit

Tests/IgniteTesting/Extensions/Date-ISO8601.swift
jaywardell: What other points of failure are you thinking about? The sut is a Date, so we don't have to worry about poorly formatted strings or the like.

I'll submit with your recommended changes, but I'm happy to add more test cases if you want.

jaywardell:
I've added a new pull request that comes from a dedicated branch rather than my main branch.
Sorry for the confusion.

See https://github.com/twostraws/Ignite/pull/383

So I'm closing this one.

PR #372, opened 2025-01-27
added tests for kebab case

jaywardell:
I broke the tests down into sets of strings that will be affected by kebabCased() in different ways.
For example, I start with the group of Strings that are not changed by the method.

I added a very minimal Instance type which just stores an input and an expected output, in order to make it easier to describe more complex conversions. idk what this type is usually called, but Instance describes its use well in my mind.

I ended with a group of kitchen-sink cases that combine two or more of the alterations that the method can perform.


twostraws:
Great job – thank you!

Issue #348, opened 2025-01-26
MetaTag properties use a bare string for property names


wwake:
From PR #344 
Paul suggested that these be changed to an enum or option set, something to enforce using one of the standard (W3C?) names.

JPToroDev:
Addressed in #373.

PR #373, opened 2025-01-27
 Add `enum` for `MetaTag` properties and names outlined in #348.

JPToroDev:


Sources/Ignite/Elements/MetaTag.swift
```diff
@@ -10,14 +10,54 @@ import Foundation
...
+        case generator
+        /// The viewport configuration for responsive web design.
+        case viewport
+
+        /// The meta attribute key.
+        var key: String { "name" }
+    }
+
+    /// Metadata properties for social media sharing.
+    public enum Property: String, Sendable {
```
twostraws, line 37: The distinction between `Name` and `Property` seems like a bagatelle, and means folks need to select the correct initializer in order to see code completion for the things they care about. Could we merge all these into a single enum, then use a computed property to return the `key` string based on a `switch`?

Sources/Ignite/Elements/MetaTag.swift
```diff
@@ -10,14 +10,54 @@ import Foundation
...
+        /// The description to display in Twitter cards.
+        case twitterDescription = "twitter:description"
+        /// The software used to generate the document.
+        case generator
+        /// The viewport configuration for responsive web design.
+        case viewport
+
+        /// The meta attribute key.
+        var key: String { "name" }
+    }
```
MrSkwiggs, line 34: I see the introduction of some more public-facing enums and decided to create #380. I'd like to have a conversation there to decide whether or not this is something we want to do instead

Sources/Ignite/Elements/MetaTag.swift
JPToroDev: Excellent idea! Will revise 🙌🏼

twostraws:
Perfect – thank you!

PR #374, opened 2025-01-27
Add tests for escaped for javascript

jaywardell:
I broke the tests up into different classes of String based on how the method would alter them.

I started with the group of Strings that will not be altered.

I then took each basic replacement in turn and added tests for them.

I used an Instance type as a shorthand for representing the input to the method and the expected output.

I ended with a group of more complex tests that combine different modifications that the method will make on a string.


JPToroDev:
Excellent—thank you!

PR #376, opened 2025-01-27
added tests for truncatedHash

jaywardell:
I printed out results of calls to String.truncatedHash to the console and then copied them into the source as arguments for test methods.

There is one test which presents a set of simple Strings that you would find in many apps (empty string, hello world, etc).
There is another one which is made up of lines from Lewis Carroll poems, which are in the public domain. The poems were taken from www.poetryfoundation.org.

In order to make the tests easily readable, I used an Instance type as shorthand to express input and expected output.


twostraws:
Thank you!

PR #379, opened 2025-01-27
Added tests for a String extension

Aradroid:
- Made it as exhaustive as I could for now
- Passed swift test and swiftlint

### DIFF
--- a/Tests/IgniteTesting/Extensions/String-CSStoJS.swift
+++ b/Tests/IgniteTesting/Extensions/String-CSStoJS.swift
@@ -175,14 +175,4 @@ struct StringCSStoJSTests {
         // Then
         #expect(jsName == "")
     }
-
-    @Test("CSS-JS conversion for a nil string")
-    func convertingCSSNamesToJS_forNilString() async throws {
-        // Given
-        let cssName: String? = nil
-        // When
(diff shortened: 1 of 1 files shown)


Tests/IgniteTesting/Extensions/String-CSStoJS.swift
```diff
@@ -14,8 +14,175 @@ import Testing
...
+    }
 
+    @Test("CSS-JS conversion for a nil string")
+    func convertingCSSNamesToJS_forNilString() async throws {
+        // Given
+        let cssName: String? = nil
+        // When
+        let jsName = cssName?.convertingCSSNamesToJS()
+        // Then
+        #expect(jsName == nil)
```
MrSkwiggs, line 186: This doesn't really test the `convertingCSSNamesToJS()` logic but rather the implementation of Swift's `Optional` type (because if `cssName` is `nil`, the compiler will never call `convertingCSSNamesToJS()`).

I think the other tests cover quite a few cases, and they do so quite well. In this case however I do not believe this test is relevant or particularly useful

Aradroid: @MrSkwiggs, thanks for explaining. I've removed it.

twostraws:
Thank you – great job!

PR #382, opened 2025-01-27
Added tests for Column element

Rimesh:
Tests for Column element
1. Basic test for Column
2. Column with ColumnSpan
3. Column with VerticalAlignment


twostraws:
Thank you!

PR #383, opened 2025-01-27
Add tests for date methods

jaywardell:
I used a Instance type to represent input and expected output.

For each property, I generated a list of ten Instance objects using random dates and then compared the output of the property to the expected output in the Instance.


jaywardell:
The way my computer calculates timezones is different from how the test server handles them, so I will remove the rfc822 code and resubmit the other.

PR #384, opened 2025-01-27
added tests based on known output

jaywardell:
I used an Instance type to keep track of input and associated expected output.

I generated strings for 10 random dates and created Instance values for each.

I iterated over them and ensured that the output is as expected.


jaywardell:
bad name sorry. will resubmit

PR #386, opened 2025-01-27
Add tests for rfc822

jaywardell:
I generated 10 random dates and expected output for each of them from Date.asRFC822.

I then encountered an issue:
RFC822 includes time zone information, but time zone of course is dependent on location.

DateFormatter.string(from:) defaults to using the timezone of the computer on which it's being run. So, as Date.asRFC822 is currently written, there is no way to create a set of tests that use an expected output string unless the computer on which the test is being run is in the same timezone as the test server. Either the tests will pass on the development machine and fail on the server, or the tests will fail on the development machine and pass on the server.

This also means that Date.asRFC822 is non-deterministic because it's location-dependent, which could be an issue in the future.

There are two possible solutions:
1) convert Date.asRFC822 to a function that takes a timezone value
2) have Date.asRFC822 use a standard timezone at all times.

I opted for the second approach and modified Date.asRFC822 to always use GMT. This means that all output will end with "+0000".

I then could use the output of the modified Date.asRFC822 to generate 10 Instance values that could be tested against consistently on any computer.

### DIFF
--- a/Sources/Ignite/Extensions/Date-RFC822.swift
+++ b/Sources/Ignite/Extensions/Date-RFC822.swift
@@ -12,6 +12,7 @@ extension Date {
     public var asRFC822: String {
         let formatter = DateFormatter()
         formatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss Z"
+        formatter.timeZone = TimeZone(secondsFromGMT: 0)
         return formatter.string(from: self)
     }
 }
--- a/Tests/IgniteTesting/Extensions/Date-RFC822.swift
+++ b/Tests/IgniteTesting/Extensions/Date-RFC822.swift
@@ -36,9 +36,3 @@ struct DateRFC822Tests {
         #expect(instance.input.asRFC822 == instance.expected)
     }
 }
-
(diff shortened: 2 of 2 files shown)


twostraws:
These tests are excellent, but I'm not sure about forcing the date to always be GMT – I can imagine some folks wanting to specify their own time zone. How about if we make the time zone a property of the `Site`, defaulting to GMT? That should allow users to have flexibility if they want it, but also ensure we can have a fixed test bed.

twostraws:
…………and I clicked the merge button by accident 😬  Apologies for the extra hassle, but I do think we should investigate the time zone thing.

jaywardell:
no hassle at all. I'll just hold onto that cell in the spreadsheet 😉

I understand your thoughts about someone wanting to use a different TimeZone. I think it would be a larger change than I'm willing to take on right now, being new to the project.

How about I convert asRFC822 to a function that takes an optional TimeZone, defaulting to GMT?  Then folks with a higher pay grade can make architectural decisions.

I'll try that out and submit in a new PR

PR #387, opened 2025-01-27
added a test to ensure that words separated by dashes are lowercased.

jaywardell:
This is a case that I hadn't considered explicitly before.
I added a test to cover this case


JPToroDev:
Awesome addition—thanks!

PR #389, opened 2025-01-27
Added test for Group element

Rimesh:
Tests for Group element
1. Group does not add any additional HTML
2. Group applies additional attribute to all its children


JPToroDev:
Looks great—thanks!

PR #391, opened 2025-01-27
Makes invoking `swiftling` more convenient

naoufalmed:
With this change, running swiftLint from the terminal will no more require specifiyng the folders to run on. Running `$ swiftlint` from the terminal will automatically only run the linter on the `Sources` and `Tests` folders.

- Moved the `.swiftLint.yml` to the route folder.
- Added an inclusion rules to scoupe linting to only `Sources` and `Tests` folders.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -14,3 +14,10 @@ identifier_name:
     - id
     - to
     - up
+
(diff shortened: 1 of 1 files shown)


twostraws:
Thank you!

PR #394, opened 2025-01-27
Stabilize ID used in animation class names.

JPToroDev:
The class names generated by the `animation()` and `transition()` modifiers now use the `Animatable` object's ID rather than the `HTML` element's internal ID. The stabilized class name will eliminate noise among diffs.

Making `Animatable` conform to `Hashable` ultimately wasn't necessary, but I've left the conformance in to make the protocol more flexible going forward.


twostraws:
Great job!

PR #396, opened 2025-01-27
Add tests for string slug

jaywardell:
I added tests for nil cases (when the method returns nil),
null cases (when the method returns its input string),
and simple cases like when a title case word is converted to lowercase.

I added tests to verify transliteration from foreign languages, stripping of emoji, stripping of diacritics and stripping of punctuation.

I added tests for replacing whitespace with dashes in simple cases (ie where there are no capital letters or other transformations).

I added tests for conversion to dash-case, and encountered several special cases such as capitals in acronyms getting converted to lowercases preceded by two dashes (where one dash might be expected).

I added tests for dealing with URL strings and file paths, two special cases that will probably be used in the project.

I added tests for dealing with simple mathematical expressions out of curiosity and to document the behavior in case it ever comes up.


twostraws:
This is incredible work – thank you!

PR #399, opened 2025-01-28
Added tests for a String extension

Aradroid:
- Some tests for `String.strippingTags()`
- swiftlint + tests passed


Aradroid:
Saw some PublishingContext Error something after I already passed all the test. Going to retry with another pull request.

PR #400, opened 2025-01-28
Added tests for extensions

Aradroid:
- Tests for String extension: `String.strippingTags()`
- Tests for URL extension: `URL.relative(to: URL)`
- Swiftlint + tests cleared


twostraws:
Thank you – this is really excellent work!

PR #406, opened 2025-01-28
Content: Fix Tables missing bootstrap table class

MrSkwiggs:
I have a md table in some of my content, and noticed it doesn't nicely layout its contents when rendered. Some quick googling led me here to the official [Bootstrap Table Documentation](https://getbootstrap.com/docs/4.0/content/tables/#examples) which states that tables are not handled by Bootstrap unless they have the `table` class.

So I've added this (along with the `scope="col"` attribute for table headers)


twostraws:
Good catch – thank you! Could you update the tests as well?

MrSkwiggs:
Done, although it seems that locally, Opacity tests are failing due to localized number formatting (generating `opacity: 0,5` instead of `opacity: 0.5`). Something we should definitely fix! Will open another PR for it

twostraws:
`0,5`: I wonder if that's some kind of localization issue?

PR #408, opened 2025-01-28
Fix number formatting

MrSkwiggs:
I wonder if there are many other places where we use the user's localized number formatting instead of a static `en-US` format. Will double check the codebase and report back


MrSkwiggs:
Only places I could find which used the locale-dependant call to `.format()` where in Slide.swift, LineSpacing.swift & Opacity.swift

JPToroDev:
Instead of requiring the use of `NumberFormatter`, could we create a computed property like `oneDecimalPlace` in extensions for `Percentage`, `Double`, and `Float`?

MrSkwiggs:
> Instead of requiring the use of `NumberFormatter`, could we create a computed property like `oneDecimalPlace` in extensions for `Percentage`, `Double`, and `Float`?

I've implemented it as an extension on FormatStyle directly, much cleaner this way. Also took the liberty of using up to 3 decimal places for opacity, but kept the rest single-decimal precision. Could make the default 3 decimal places always?

JPToroDev:
> I've implemented it as an extension on FormatStyle directly, much cleaner this way.

Awesome!

Could we make the extension a little more concise? nonLocalizedDecimal(_ decimalPlaces: Int = 1) ?

> Could make the default 3 decimal places always?

I like the way you have it 🙂

MrSkwiggs:
> Could we make the extension a little more concise? nonLocalizedDecimal(_ decimalPlaces: Int = 1) ?

`value.formatted(.nonLocalizedDecimal(3))` isn't a super obvious callsite in my opinion :/ 
Maybe a good compromise would be `value.formatted(.nonLocalizedDecimal(precision: 3))` makes it a bit more concise while still being clear?

Let me know 🙏

MrSkwiggs:
Also I see that we might be having a race condition since the changes to the PublishingContext were made, as last tests failed with `Ignite/PublishingContext.swift:22: Fatal error: "PublishingContext.default accessed before being initialized. Call PublishingContext.initialize() first`. Running tests locally a few times I could reproduce this as well. Will make an issue for it 👌

JPToroDev:
> Also I see that we might be having a race condition since the changes to the PublishingContext were made, as last tests failed with

Yes, the way we're currently approaching tests doesn't play nice with `PublishingContext`, but #420 sorts this all out 🙌🏼

MrSkwiggs:
> How about `length` or `places` then? Would shave off a few more character

Done, using `places` 👍

JPToroDev:
Excellent—thank you!

PR #409, opened 2025-01-28
Convert rfc822 to a function

jaywardell:
I converted Date.asRFC822 to a function that takes a TimeZone, defaulting to GMT.

This way, client code can specify what time zone the output string should represent.

IF timezone is added as a parameter on Site, then the Site's time zone could be passed to Date.asRFC822(timeZone:).

I had to make a few changes to client code so that it calls the function rather than simply requesting the property.

I also added several tests against different time zones.

Sadly, the way Foundation does time zone calculations for historical dates is beyond my understanding, so I made the choice to only test dates after January 1, 1970. I don't think this will be a problem for Ignite, unless someone is going to go back in time and publish a website...


JPToroDev:
Great work!

> IF timezone is added as a parameter on Site, then the Site's time zone could be passed to Date.asRFC822(timeZone:).

Is this an enhancement you'd be interested in contributing?

jaywardell:
As I told Paul in https://github.com/twostraws/Ignite/pull/386, I'm new to the project's architecture and idk if a change like that would have more implications than anticipated. 

A quick search of the code base tells me that right now the implications are only to FeedGenerator.swift (Ignite) and FeedGenerator.swift (Testing). 

If it's just a matter of a couple lines of code, sure. But could it affect anything besides those 2 modules?

JPToroDev:
> A quick search of the code base tells me that right now the implications are only to FeedGenerator.swift (Ignite) and FeedGenerator.swift (Testing).

Yes, from what I can see, that's the only code involved.

> I'm new to the project's architecture

We're all new until we're not, and you've been doing great work so far.

If adding this feature to `Site` excites you, I'm more than happy to help you through your first enhancement contribution. Of course, no pressure either way!

jaywardell:
ok. I'll give it a go in another pull request.

jaywardell:
I just submitted pull request: https://github.com/twostraws/Ignite/pull/415 to add a timezone property to Site.
It depends on changes from this one, but I submitted it separately in case this one checks out but that one needs more work.

JPToroDev:
Great! I'll be making some code-level comments over there. In the meantime, I can merge this one in once the conflicts are resolved.

PR #414, opened 2025-01-28
Fix  occasional race conditions when testing

jaywardell:
When testing in Xcode, I was encountering an occasional crash that was happening when other tests were being run (tests other than mine).  

Screenshot 2025-01-28 at 10 24 31 AM

They appear to be related to these tests not calling PublishingContext.initialize() before calling render(), though actually there are a few other tests that don't call PublishingContext.initialize() before calling render() and don't seem to cause this issue.

This behavior can be reproduced pretty reliably by control-clicking the test button on line 14 of IncludeTests in Include.swift and choosing "Run 'Include Tests' Repeatedly…"
Run 'Include Tests'

As I say, there are a few tests that can cause this crash, and it's flakey exactly when it will happen. 

Still, by calling `try PublishingContext.initialize(for: TestSite(), from: #filePath)` on init, this crash can be avoided. This is at least the approach that others have used in tests like SubsiteLinkTests in Link.swift.

So I added this to a few tests that seem to cause this crash for me. If others encounter this with other tests, they can feel free to copy this approach.


twostraws:
Thank you! We've used this `init()` solution elsewhere, and it's really flaky – so easy to forget, and so damaging. While we think up a better solution we'll need to proceed with this. Did it pass SwiftLint okay?

JPToroDev:
I've got a fix to this janky `init()` hack we've been using. Will push shortly.

jaywardell:
I thought that I ran it through swiftlint, but I can’t say for sure, and now I’m afk at the 9 to 5. 

I’ll recheck when I get home. 

Sorry for the bother.

jaywardell:
I had forgotten to run SwiftLint on this PR. Sorry.

It looks like this is probably moot given https://github.com/twostraws/Ignite/pull/420. Feel free to close this if it's accepted.

JPToroDev:
I'm going to close this PR since #420 introduces a long-term solution. Excited to get your other PRs across the finish line!

jaywardell:
👍

PR #415, opened 2025-01-28
Add timezone parameter to Site protocol

jaywardell:
I added a timeZone property to Site.
It defaults to nil.
The expectation is that code should assume a GMT time zone if a Site's timeZone is nil.

FeedGenerator.generateContentXML() outputs the date string appropriate to the Site's time zone in the <pubDate> tag.

TestSite was modified to optionally have a custom timeZone in order to be able to test code that depends on a Site's timezone.

FeedGeneratorTests.generateFeed() takes a parameterized list of Sites to test against, including ones with custom time zones.


twostraws:
Thank you! If we're defaulting to GMT, would it be sensible to make the property use GMT as its default value rather than being optional? If someone sees `nil` as the default value, it asks more questions than it answers 🙂

JPToroDev:
I see. So if `nil` is a valid value, could we keep it as an optional, but still make GMT the default value?

jaywardell:
I've added a .gmt static var to TimeZone and set that as the default value.
Honestly, I don't love it. In order to explain the logic, here's the current comment:

    /// the time zone used to do date output for your site. Defaults to GMT
    /// (nil is equivalent to GMT)
    var timeZone: TimeZone? { get }

This seems to me like more cognitive load than is needed for something that most people won't need or want to change.

JPToroDev:
After looking through the `DateFormatter` documentation, I see the confusion.

`DateFormatter` doesn't default to GMT, we do. 

We should follow how `DateFormatter` works: where an unspecified value (`nil` in our case) defaults to the system time zone. 

So yes, `nil` should be our default, and we should should do something like this when configuring the date formatter since its `timeZone` is force unwrapped:

```swift
if let timeZone {
    dateFormatter.timeZone = timeZone
}
```

jaywardell:
defaulting to gmt was the goal of the PR as I understood it.
I was responding to @twostraws idea in https://github.com/twostraws/Ignite/pull/386.

Also, allowing the formatter's timezone to be nil in `asRFC822()` causes the base case to fail in DateRFC822Tests.DateRFC822Tests, which was the impetus for this whole thing. 🔃😂

Honestly, I think it's probably okay to default to a different value than `DateFormatter` does. The behavior is well-documented and the folks who would care can figure it out. 

There's also a minor privacy concern to consider: If we publish the user's timezone in the source (like `DateFormatter` does when its `timeZone` is nil), we're providing location info about the user that they may not realize we're providing. Having a consistent default of GMT prevents that possibility. If the user wants to publish that info, they can override their own`timeZone` parameter in their custom `Site`. By default, their site can have an industry standard timezone in its `pubDate` and no-one thinks anything of it.

Maybe we don't allow a nil time zone at all? Seems heavy-handed but it would make most of this discussion moot.

jaywardell:
> Maybe we don't allow a nil time zone at all? Seems heavy-handed but it would make most of this discussion moot.


...and when I say this, I mean we make the `timeZone` property non-optional, we make the default `.gmt` and we let the app crash if the user overrides the property and accidentally returns a nil value (e.g. if they use `init(abreviation:)` and pass a value that Foundation doesn't accept.

JPToroDev:
I see, all great points. Would it be possible to keep it optional with GMT as the default and allow the user to use `nil` if they wanted without any crashes or issues? That would be optimal and would mean the PR is ready to go as is, right?

Thanks for the great work on this!

jaywardell:
That should be the PR as is. In fact, most users will probably not even know that they’re using nil since it’s the default.

JPToroDev:
Perfect! I'll merge this in as soon as I'm back at the computer 🙌🏼

PR #416, opened 2025-01-28
Added test for AspectRatio Modifier - including commented out bug test

TheApApp:
Added a series of tests for the Aspect Ration Modifier.

### DIFF
--- a/Tests/IgniteTesting/Modifiers/AspectRatio.swift
+++ b/Tests/IgniteTesting/Modifiers/AspectRatio.swift
@@ -14,8 +14,80 @@ import Testing
 @Suite("AspectRatio Tests")
 @MainActor
 struct AspectRatioTests {
-    @Test("ExampleTest")
-    func example() async throws {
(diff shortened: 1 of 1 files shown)


Tests/IgniteTesting/Modifiers/AspectRatio.swift
```diff
@@ -14,8 +14,80 @@ import Testing
...
+        let expectedValues: [ContentMode] = [.fit, .fill]
+
+        let expectedRawValues = expectedValues.map { $0.htmlClass }
+        let actualRawValues = ["object-fit-contain", "object-fit-cover"]
+
+        #expect(expectedRawValues == actualRawValues)
+    }
+
+    @Test("Verify .square")
+    func verifySquareModifier() async throws {
```
twostraws, line 38: Can this and the subsequent tests become parameterized?

twostraws:
I'll merge this in now, but I think we should return to it again to pass any values into Swift Testing's macro. Thank you!

PR #420, opened 2025-01-28
Fix janky tests by introducing `IgniteTestSuite`

JPToroDev:
In the Swift Testing documentation, it says if you need set-up and tear-down behavior in your tests, it's perfectly valid to make the suite a class instead of a struct. Accordingly, I've added `IgniteTestSuite` and made all applicable tests a subclass, with the `init` and `deinit` handling the publishing context lifecycle.


twostraws:
This seems like the best solution, although `BaseIgniteTest` or similar would work better than anything starting UI, only to avoid problems with folks with a UIKit background. Also, before we merge this, we need to merge the other outstanding test PRs, so everything is up to date before we migrate across.

Tests/IgniteTesting/SiteDependent.swift
```diff
@@ -0,0 +1,32 @@
...
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+import Foundation
+@testable import Ignite
+
+@MainActor
+@propertyWrapper
+public struct SiteDependent<Value: HTML> {
```
twostraws, line 13: I'm not sold on this approach just yet. Swift Testing's parallel approach makes me think this clears up some of the complexity, but not all of it. I'd be tempted to bring in the class-based testing approach, then push back subsite testing until we have a better solution – perhaps one as simple as putting all subsite tests in their own, isolated class.

JPToroDev:
> This seems like the best solution, although BaseIgniteTest or similar would work better than anything starting UI, only to avoid problems with folks with a UIKit background.

I originally had `TestSuite`. The reason I added the `UI` prefix is because we don't actually need to use that subclass for all our tests and that seemed like a clear deliminator. But I agree it can be misleading/confusing.

If we do go with something like `BaseIgniteTest`, it'd probably be best to just apply that to every test suite to eliminate any confusion.

Tests/IgniteTesting/SiteDependent.swift
JPToroDev: That works for me. So should we remove `@SiteDependent` now, or leave it as a band-aid until we isolate the subsite tests?

JPToroDev:
Should be good to go now. Let me know if you'd like anything tweaked.

twostraws:
Perfect!

PR #427, opened 2025-01-29
Added tests for extensions on Array and Sequence

Aradroid:
- Tested against simple custom type
- Tested against simple custom type conforming to HTML protocol


twostraws:
Thank you! Really comprehensive 🙌

PR #431, opened 2025-01-29
Background modifier tests

olofhennig:
Starting with tests for Color, String color, Material and Gradient.


JPToroDev:
Great thank you!

PR #432, opened 2025-01-29
Refined the Color test suite from the initial workshop attempt

bobdel:
This is my second pass on making the Color tests. It still requires work but I have learned a lot already.

### DIFF
--- a/Sources/Ignite/Framework/Content.swift
+++ b/Sources/Ignite/Framework/Content.swift
@@ -200,7 +200,7 @@ public struct Content {
         let formatter = DateFormatter()
         formatter.timeZone = .gmt
 
-        let formats = ["y-M-d HH:mm", "y-M-d"]
+        let formats = ["y-M-d", "y-M-d HH:mm", "y-M-d H:m", "y-M-d HH:mm:ss", "y-M-d H:m:s"]
         for format in formats {
             formatter.dateFormat = format
             if let date = formatter.date(from: date) {


Tests/IgniteTesting/Types/Color.swift
```diff
@@ -11,21 +11,46 @@ import Testing
 @testable import Ignite
 
 /// Tests for the `Color` type.
-@Suite("Color Tests")
+@Suite("Color Type Tests")
```
NigelGee, line 14: Is this test already on `foregroundStyle` test?

bobdel: That is possible. I'll look for a conflict. I used the .foregroundStyle modifier in the initial run of testing Color, before realizing that was an unnecessary dependency on the Color texts. This commit included progress toward removing it altogether, which my next goal. The main goal for creating this commit was to acknowledge that the first version wasn't that good, as a placeholder until I get it right.

NigelGee: What I was saying that the test you had there was the same as the test in Foreground Modifier so it just a repeat of it. Thinking is to test the different init like
```
@Test("Color Hex")
  func colorName() async throws {
      let colorHex = Color(hex: "#F0F8FF")

      let output = colorHex.description

      #expect(output == "rgb(240 248 255 / 100%)")
  }
```
Just up to you

bobdel:
Thanks, Nigel. Now I understand. I'm working on a revision now.

PR #430, opened 2025-01-29
Tests for the `Background` modifier.

NigelGee:
Change to `Gradient`, `GradientType`, `Material` and `ColorScheme` to make them `Sendable`. Hope my formatting is okay this time.

PR #437, opened 2025-01-30
Update for Background Tests.

NigelGee:
I have change the Material and Gradient ones as to #430. Left Color and String ones

### DIFF
--- a/Sources/Ignite/Styles/Gradient.swift
+++ b/Sources/Ignite/Styles/Gradient.swift
@@ -6,9 +6,9 @@
 //
 
 /// A gradient that transitions between colors.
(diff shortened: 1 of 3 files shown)


NigelGee:
Just on a footnote. If a Material of say `@MainActor public static let testMaterialColor = Material(type: .thin, colorScheme: .dark)` was added to `Material` then this could be add to test the whole of `className` and I could not add it to just the test because private init.

JPToroDev:
There's a whole suite of tests dedicated to `Material` itself, so what you have should be more than enough. Great work!

NigelGee:
Yes I saw that there was so will leave it as is

PR #438, opened 2025-01-30
Add SiteTests

erikbasargin:
This PR addresses the `"PublishingContext.default accessed before being initialized. Call PublishingContext.initialize() first.` error when Markdown content contains invalid lastModified.

* `Content/parseMetadataDate` is updated to set error on the provided context rather than shared instance of `PublishingContext`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 Package.resolved
 /.build
 /Packages
+Tests/IgniteTesting/TestWebsitePackage/Build
+Tests/IgniteTesting/TestWebsitePackage/Content/*.md
 xcuserdata/
 DerivedData/
 .swiftpm/configuration/registries.json
--- a/Tests/IgniteTesting/Publishing/Site.swift
+++ b/Tests/IgniteTesting/Publishing/Site.swift
@@ -1,0 +1,83 @@
+//
+// Site.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+import Foundation
+import Testing
+
+@testable import Ignite
+
+/// Tests for the `Site` type.
+@Suite("Site Tests", .serialized)
+@MainActor
+struct SiteTests {
+    
+    private let package = TestPackage()
+    
+    @Test("Site published given there is no Markdown content")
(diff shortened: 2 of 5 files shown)


JPToroDev:
Please see my note about `parseMetadataDate()`—those changes should be undone—otherwise this looks excellent!

JPToroDev:
You've add a `context` argument to that method that wasn't there before, right?

JPToroDev:
That does solve the problem in that one spots, but `parseMetadataDate()` is a symptom of a larger issue that #420 addresses. 

So we should undo that change and just focus on the tests. Even if they fail at the moment, once #420 is merged in, it'll be resolved naturally.

erikbasargin:
I probably missing something, I can update this PR just to provide `SiteTests`. However, I want to point out that `SiteTests/publishingWithInvalidLastModifiedDate` a real life scenario, I mean `ignite build` failed for me yesterday when I tried to build a website. 

#420 seems only covers IgniteTesting.

```
ignite build
⚙️  Building your site...

Building for debugging...
[0/3] Write swift-version--58304C5D6DBC2206.txt
Build of product 'MyBlog' complete! (0.11s)
Ignite/PublishingContext.swift:22: Fatal error: "PublishingContext.default accessed before being initialized. Call PublishingContext.initialize() first.
```

JPToroDev:
I see, thank you for clarifying. Could you open up a separate issue outlining the Ignite build failure you experienced yesterday that's separate from testing?

erikbasargin:
I rebased my tests on top of your changes #420 (https://github.com/erikbasargin/Ignite/tree/012825a), and you're absolutely right that tests are fixed. However, `ignite build` still fails when `lastModified` date has invalid format in Markdown files. I'll check how I can update my tests so they would still cover the issue in `parseMetadataDate` function including #420 changes. 

Do you want to merge current SiteTests? I can raise another PR a bit later today with proposed solution.

JPToroDev:
When the build fails, do you get an error like "Content should have a date in the format of XYZ", or the publishing context `nil` error?

JPToroDev:
Perfect—thank you. Include this all in your new PR, along with your proposed solution using the revamped `parseMetadataDate()`, and we'll get this resolved shortly.

Issue #445, opened 2025-01-31
[Bug] "Ignite build" crashes when Markdown content contains invalid lastModified date


erikbasargin:
I tried to build a website with the `ignite build` command, and haven't noticed the `lastModified` date with unsupported format in one of the Markdown files, e.g. `2020-03-30 16:37:21`.

## Problem 

```
ignite build
⚙️  Building your site...

[0/1] Planning build
Building for debugging...
[0/3] Write swift-version--58304C5D6DBC2206.txt
Build of product 'MyBlog' complete! (0.18s)
Ignite/PublishingContext.swift:22: Fatal error: "PublishingContext.default accessed before being initialized. Call PublishingContext.initialize() first.


❌ Failed to generate HTML.
```

## Expected behaviour

```
ignite build
⚙️  Building your site...
📘 Publish completed with exceptions:
	📕 Content dates should be in the format 2024-05-24 15:30.


✅ Successfully built!
```

JPToroDev:
Thank you! I think I've pinpointed the true source of this issue in #448. Does that fix the issue on your end?

erikbasargin:
Yes, it works! `SiteTests/publishingWithInvalidLastModifiedDate` test is pass with #448 changes.

I'm going to remove my fix from #445, there is still some SiteTests updates that increase stability of the tests, and a note that point to a bigger issue (not related to #445).

PR #446, opened 2025-01-31
Increased stability of SiteTests

erikbasargin:
This PR addresses #445 issue, and continuation of #438.


erikbasargin:
The fix has been removed in favour of #448. Here are still some SiteTests updates that increase stability of the tests, and a note that point to a bigger issue with depending on `PublishingContext.default` across unit tests.

PR #450, opened 2025-01-31
Add tests for NavigationBar

jaywardell:
I added a few methods on String in the test target that make searching for html tags and attributes simpler (in a testing environment only, these obviously aren't robust enough for general work)

Using these methods, I added tests for the structure of the output of NavigationBar as well as for the various parameters that can be applied to NavigationBar.

I had to make one change to NavigationBar.Width, making it Sendable so that it can be parameterized in the tests.

idk, maybe the String extensions should be in a separate PR, but I thought that providing an example of their usefulness would be appropriate. Please advise


twostraws:
Thank you!

Issue #451, opened 2025-01-31
[Xcode 16.2] LengthUnitTests/isDefault crashes with "Internal inconsistency" error


erikbasargin:
When I run `LengthUnitTests/isDefault` locally I get a crash. I'm using Xcode 16.2.

```
◇ Suite "LengthUnit Tests" started.
◇ Test "Test is default value." started.
◇ Passing 2 arguments unit → 20.0em, expected → false to "Test is default value."
◇ Passing 2 arguments unit → infem, expected → true to "Test is default value."
◇ Passing 2 arguments unit → infem, expected → true to "Test is default value."
◇ Passing 2 arguments unit → inf%, expected → false to "Test is default value."
​​​​✔ Test "Test is default value." passed after 0.001 seconds.
✔ Suite "LengthUnit Tests" passed after 0.001 seconds.
✔ Test run with 1 test passed after 0.001 seconds.
XCTest/HarnessEventHandler.swift:282: Fatal error: Internal inconsistency: No test reporter for test case argumentIDs: nil in test IgniteTesting.LengthUnitTests/isDefault(unit:expected:)/LengthUnit.swift:101:6
```

JPToroDev:
Thank you!

PR #452, opened 2025-01-31
Fix LengthUnit/isDefault test crash on Xcode 16

erikbasargin:
Fix for #451 issue.

This seems to be a bug in Xcode 16. See swift forum discussion for more details https://forums.swift.org/t/fatal-error-internal-inconsistency-no-test-reporter-for-test-case-argumentids/75666.

The test crashes because there are two copies of the same value `LengthUnit.em(.infinity)/LengthUnit.default`. Moved the check of LengthUnit.default into a separate test `LengthUnitTests/defaultValue` to resolve this crash.


JPToroDev:
Excellent—thank you!

PR #454, opened 2025-02-01
Added a test for Row

EdgedSword:


EdgedSword:
Thanks and yes both of these files passed SwiftLint. 

I did see that the IgniteTesting/Publishing/Site.swift and IgniteTesting/TestWebsitePackage/Sources/TestSite.swift had linting violations. I didn't know if those were in work so I left them but I can clean those up if you would like?

JPToroDev:
I just pushed a big cleanup PR!

JPToroDev:
Thanks for the great work!

PR #458, opened 2025-02-01
Eliminate `PublishingError` in favor of simple strings, mirroring `warnings` and ensuring fatal errors crash at the call site.

JPToroDev:
An improvement on our working error-handling system until we officially tackle the subject.

### DIFF
--- a/Sources/Ignite/Extensions/URL-SelectDirectories.swift
+++ b/Sources/Ignite/Extensions/URL-SelectDirectories.swift
@@ -25,7 +25,7 @@ extension URL {
             }
         }
 
-        throw PublishingError.missingPackageDirectory
+        fatalError(.missingPackageDirectory)
     }
 }
 
--- a/Sources/Ignite/Framework/Site.swift
+++ b/Sources/Ignite/Framework/Site.swift
@@ -264,7 +264,7 @@ public extension Site {
 
         if !context.warnings.isEmpty || !context.errors.isEmpty {
             print("📘 Publish completed with exceptions:")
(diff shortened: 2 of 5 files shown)


wti:
I don't understand the motivation, but noticed some possible interactions with content symlinks:

1. Content symlink (#264) supports configurable number of errors to tolerate, avoiding fatal errors.  The goal is to enable the user to tolerate some content/parsing errors instead breaking the whole site build, on the assumption that it's better to omit the invalid content.
2. I noticed a "duplicate directory" error and wondered if content-symlink code should be throwing that instead of the one it does.  

Having some common enum error types for related errors (e.g., parsing vs assembly vs resource-discovery) with a number of cases would simplify error-handling (perhaps the user could handle and ignore cases at will) and help with discoverability (as here with my finding the existing error).  As an interim measure if it's necessary to enforce immediate fatal error in some cases, you could use a registery for enums or swap struct for enum and use an initializer.  

The approach of consolidating error creation and handling seems robust to the frequent addition of new errors.   But my sympathies in any case; error handling is pervasive and extensive, and it's hard to simplify :)

JPToroDev:
This PR was meant to make the current behavior better without fully reworking error and warning handling (which is needed and will be done down the line), but there are much more pressing PRs, so I'll close this for the time being.

JPToroDev:
I've added some comments throughout the PR. Overall the code looks good, but a lot of the naming isn't very reader friendly. Once those notes are addressed, I'll be very happy to merge this in—let me know if you have any questions.

JPToroDev:
Sorry—hit "Close with comment" instead of "Comment" 🤦🏻

PR #462, opened 2025-02-05
Add initializer to `Section` that takes a header argument and wraps its children in a `<section>` instead of a `<div>`.

JPToroDev:
It's likely that some users might expect Ignite's `Section` element to wrap its children in a `<section>`, not a `<div>`. This PR empowers `Section` to do just that depending on context (the initializer used).

### DIFF
--- a/Sources/Ignite/Elements/Section.swift
+++ b/Sources/Ignite/Elements/Section.swift
@@ -5,14 +5,13 @@
 // See LICENSE for license information.
 //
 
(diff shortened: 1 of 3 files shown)


twostraws:
I've thought about this a lot, and the surprise tag change still feels odd. But at the same time, I think it's the natural result of a declarative layout – we're saying we want a section, and whether that becomes a `<section>` or a `<div>` is an implementation detail users really ought not to worry about. So, TL;DR: this is fine 😅

PR #470, opened 2025-02-06
moved test to appropriate module

jaywardell:
This one test that was in Testing/IgniteTests/Eleemnts/HTMLHead.swift tests a method on MetaLink, not a method on HTMLHead. While the output DOES appear in HTMLHead, it probably should be in the same module as other tests on MetaLinks.


twostraws:
Quite correct – thank you!

PR #479, opened 2025-02-07
Simplify Ignite: Eliminate `PageContext` by incorporating it into `EnvironmentValues`

JPToroDev:
`PageContext` is essentially the same exact object as `EnvironmentStore`—a singleton that creates a context for the current page being published, and infuses that context with relevant values.

This similarity gives us the opportunity to consolidate these objects and make creating layouts even simpler.

This PR is the first in a larger series, and while it contains several changes, they are all in service of simplifying this code:

```swift
struct MainLayout: Layout {
   var body: some HTML {
       HTMLDocument {
           HTMLHead(for: page)
           HTMLBody {
               NavBar()
               Section(page.body)
               Section {
                   SocialFooter()
                   IgniteFooter()
               }
           }
           .padding(.vertical, 80)
       }
   }
}
```

Into this code:

```swift
struct MainLayout: Layout {
   var body: some HTML {
       Body {
           NavBar()
           Section(page.body)
           Section {
               SocialFooter()
               IgniteFooter()
           }
       }
       .padding(.vertical, 80)
   }
}
```

`Head` and `Body` now mirror scenes in SwiftUI, and the default `Head` is automatically provided by the result builder, hence its omission above. `Head` needs to be included only if you want extra head elements like `Analytics`.

The superficial changes:
- Restored `HTMLHead` and `HTMLBody` to `Head` and `Body` by reworking the associated type names that originally confused the compiler. Renamed `HTMLDocument` to `Root`.
- Added an internal `environment` default property to `RootHTML`

The main change is adding a `page` property to `EnvironmentValues`, which we access in the initializers of `Head` and `Body` now, instead of requiring a page as an input.


Sources/Ignite/Elements/Root.swift
```diff
@@ -5,7 +5,7 @@
 // See LICENSE for license information.
 //
 
-public struct HTMLDocument: HTML {
+struct Root: HTML {
```
twostraws, line 8: I'm not sure about the name `Root`. `Document`, perhaps? Or something else?

PR #480, opened 2025-02-07
Simplify Ignite: Eliminate `ContentContext` by incorporating it into `EnvironmentValues`

JPToroDev:
This PR builds on #479. It adds a `currentContent` property to `EnvironmentValues` and eliminates the need for `ContentContext`.


twostraws:
I'm okay with bringing `ContentContext` into the environment, but this PR seems to do a whole bunch of other things?

JPToroDev:
Cool! This PR is meant to be merged in after the earlier PRs in the series, which is why it appears there are extra things mixed in.

PR #481, opened 2025-02-07
Simplify Ignite: Eliminate `TagContext` by incorporating it into `EnvironmentValues`

JPToroDev:
This PR builds on #479 & #480. It adds `currentTag` and `taggedContent` properties to `EnvironmentValues` and eliminates the need for `TagContext`.


twostraws:
My brain is really segfaulting on these PRs – they seem to be very similar, if not identical.

JPToroDev:
This PR is meant to be merged in after the earlier PRs in the series, since it relies on some of those changes, which is why it appears there are extra things mixed in.

PR #482, opened 2025-02-07
Simplify Ignite: Make `Layout` types less abstract and more intuitive

JPToroDev:
This PR builds on #479, #480, & #481.

Ignite currently has several types of "layouts"—`StaticLayout`, `ContentLayout`, `TagLayout`, and `Layout`.

These terms are abstract, and its hard to discern their purpose from their names.

On top of that, `StaticLayout`, `ContentLayout`, and `TagLayout` can have a parent layout—which is confusing because aren't they themselves layouts? Why would a layout need a layout?

(`StaticLayout`, `ContentLayout`, and `TagLayout` affect the layout of a page's *content*, while `Layout` affects the layout of the page itself—confusing!)

While these names served their purpose during the recent reengineering—and were a little more logical before the layout variant types could adopt a parent layout—they now require a little TLC.

This PR aims to eliminate this confusion with the following main renamings:

1. `StaticLayout` -> `Page`
To clearly communicate we're dealing with a page—content and content layout.

2. `ContentLayout` -> `Article`
To clearly communicate we're dealing with an article—content and content layout.

3. `TagLayout` -> `Archive`
To clearly communicate we're dealing with an archive page—content and content layout.

All of these protocols conform to `PageContent` (see #484) to clearly communicate they deal with a page's content.

Not only are these terms more concrete, familiar, and intuitive, but they're also more concise—double win!


twostraws:
I'm struggling to see the approach here, but these names don't sit comfortably with me. We already have a core type called `Content`, encapsulating a piece of content. The distinction between that and `Article` is extremely fine, and we've already extensively renamed these things in the past.

JPToroDev:
When we first devised these names, `StaticLayout`, `ContentLayout`, and `TagLayout` defined both the content of a page *and* its layout. They couldn't adopt a `Layout`—they *were* the layout. In that context, these names made perfect sense.

But after settling on these names, we reintroduced the ability for these types to specify a parent layout—which is what makes these name so confusing now.

Why is a layout defining another layout? Where does one start and the other end? And what are these two "layouts" relationship to one another?

Most folks would expect `StaticLayout`, `ContentLayout`, and `TagLayout to *conform* to `Layout`, but that's also not the case.

I know we've renamed things and that that chapter needs to come to a close, so I don't make this PR lightly, but when trying to explain how Ignite works to folks, this has been a huge point of confusion.

As for `Article` vs `Content`, I don't think the distinction is fine at all—a social-media post is content, an episode of TV is content, an article is content. But content is not an article.

PR #485, opened 2025-02-08
Tests for hidden modifier

ebariaux:
In hiddenMediaQueryForText() test, I had to set a theme on CSSManager for the className(for:) call to return the class name instead of an empty string.
That is because, in CSSManager, the register MediaQuery (called by the HiddenModifier initialiser) creates a pending registration in case there is no theme.

I could write the test using hashForQueries() instead of className(for:) but:
- I'm now leaking implementation details in my test
- what would happen when generating a site if no theme is set (sorry, I don't know Ignite enough - haven't taken the time to create a test site)


twostraws:
Thank you! I think you make the right choice 👍

PR #498, opened 2025-02-13
Added a test for fontModifier

EdgedSword:
I was not sure the proper way to approach the test Responsive Font Size Application due to the registerClasses method being private and trying to confirm the expected output. I created the Helper Function simulateRegisterClasses.

Please let me know if this works or if there is a different approach to take.


twostraws:
Well, it's definitely a start, and something we can iterate on going forwards. It also locks down our existing functionality, which is excellent. Thank you!

PR #500, opened 2025-02-14
Fix main actor isolation errors when building as Xcode target

### LINKED ISSUE #499: Main actor isolation errors
## Summary

The library builds without an issue when used as a Swift Package. However, when the source code in embedded in Xcode framework target, it can't be compiled with Xcode 16.2, due to the following error:

> ❌ Call to main actor-isolated [...] in a synchronous nonisolated context

The error occurs in couple places, and is caused by the way how Xcode performs strict concurrency checking. It happens with strict concurrency checking set to minimal, and language mode set to Swift 5 in Xcode target's build settings.

## How does it affect developers

One can't build Ignite as Xcode framework target, or create XCFramework form it. 

This can happen when using [Tuist](https://github.com/tuist/tuist) to add Ignite as a dependency. Tuist creates Xcode projects with framework targets for Swift Package dependencies. Thanks to this approach, the dependencies can be build into XCFrameworks, and cached. It also improves overall developer experience, solving issues with Xcode package resolution that slows down development.

darrarski:
## Summary

- Resolves #499

This PR contains changes that resolves the issues when building Ignite as Xcode target. For example, when Ignite is integrated as a dependency using [Tuist](https://github.com/tuist/tuist).

## What was done

- [x] Annotate the following protocols with `@MainActor`: `BlockHTML`, `InlineElement`, `RootElement`, `Layoutable`, and `HorizontalAligning`.
- [x] Annotate `EmptyHTML.init` with `nonisolated`.

## Additional info

Although the changes included in this PR solves the problem described in the issue, more changes may be required to improve actor-isolation issues. This PR contains minimal amount of changes to the source code that allows to compile Ignite when integrating it using Tuist.

### DIFF
--- a/Sources/Ignite/Elements/EmptyHTML.swift
+++ b/Sources/Ignite/Elements/EmptyHTML.swift
@@ -9,7 +9,7 @@
 /// Used as a default or fallback when no content is needed
 public struct EmptyHTML: HTML, InlineElement, RootElement {
     /// Creates a new empty HTML element
-    public init() {}
+    public nonisolated init() {}
 
     /// Returns self as the body content since this is an empty element
     public var body: some HTML { self }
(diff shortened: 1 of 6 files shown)


JPToroDev:
Thank you!

PR #504, opened 2025-02-15
Tests for horizontalAlignment modifier

ebariaux:
More tests
There were already some Swift Lint validation errors before the branch, should be addressed in separate PR.


JPToroDev:
Thank you!

PR #528, opened 2025-02-15
Restore `HTMLBody` to `Body`.

JPToroDev:
During the development of Ignite’s imminent new release, `Body` was changed to `HTMLBody` to avoid a collision with `Layout`’s associated type name. The associated type name has been reworked to permit the use of `Body` again, improving ergonomics.


twostraws:
Thank you!

PR #529, opened 2025-02-15
Restore `HTMLHead` to `Head`.

JPToroDev:
During the development of Ignite’s imminent new release, `Head` was changed to `HTMLHead` to complement `HTMLBody`. Recent changes have allowed the prior names of both types to be restored—see #528.


twostraws:
Great!

PR #530, opened 2025-02-15
Rename `HTMLDocument` to `Document`, `RootElement` to `DocumentElement`, and `RootElementBuilder` to `DocumentElementBuilder`.

JPToroDev:
This PR renames `HTMLDocument` to `Root`, creating parallelism between it and other top level types like `Head` and `Body`. This new name also creates a lexical relationship between it and the `RootElement` protocol—just like `Head` and `HeadElement`'s relationship.

`Document` is a viable alternative to `Root`, although that would require us to change `RootElement` to `DocumentElement` to match.

This PR tees up a future PR where `Root` will become an internal type.


twostraws:
`Document` has more meaning than `Root`, and aligns with the existing term "HTML document". Why would this type need to become fully internal?

JPToroDev:
The result builder will automatically add it, which means it doesn't need to be exposed.

`Document` works. Are you fine with `DocumentElement` though? 

I guess we could keep `RootElement`, but it'd be a shame for those two types not to have an obvious lexical relationship.

twostraws:
Having given this time to percolate, I think `DocumentElement` is fine, and makes the whole PR fit together neatly with the rest of the project and prior art in HTML.

PR #531, opened 2025-02-15
Rename `CoreAttributes` to `Descriptor`.

JPToroDev:
This PR addresses the longstanding issue that `CoreAttributes` has evolved to contain much more than just HTML attributes and that its name no longer accurately reflects its function.

This PR proposes the name `Descriptor` as a more comprehensive replacement. If `Descriptor` feels a bit too generic, the variant `ElementDescriptor` could work nicely as well.

This is an internal type with no breakage.


twostraws:
"Descriptor" is indeed too vague, and it confuses the Unix side me where concepts like "file descriptor" have a very specific meaning. What is the full range of things `CoreAttributes` is currently responsible for?

JPToroDev:
I was taking inspiration from Apple's `FontDescriptor` and `CoreAttribute`'s primary `description()` method—but your UNIX point is a good one. We can do better.

Here's the run down:

all HTML attributes
HTML tags
columnWidth
attributes of specialized wrapper divs

I was thinking something like `CoreMarkup` could work too, although "markup" usually implies the content of the element, too—not just its "wrapper" components—but I think the "core" prefix communicates its not comprehensive.

Do you like that more?

twostraws:
Is the issue the word "attributes"? Obviously it has a double meaning here – the HTML-specific meaning of key-value pairs inside elements, and the more general one, meaning "properties". Specification? Metadata? `TagInfo`? If none of those feel better to you, then I think we're back to `Descriptor`, or perhaps something a little more focused like `TagDescriptor`.

JPToroDev:
Yes, I think "attributes" is the misleading word here. Adding the `tag` property is when you suggested we rename it.

Metadata isn't bad, although I wonder if that name is a little confusing in context of the `MetaTag` type—or even a more standardized name for that type within Ignite.

Otherwise, I think `TagInfo` or `TagDescriptor` work nicely.

JPToroDev:
That said, because `CoreAttributes` also contains info about specialized wrapper divs, the prefix `Element` might be preferable to `Tag`. So worth considering:

`TagInfo`
`TagDescriptor`
`ElementInfo`
`ElementDescriptor`

twostraws:
Okay, let's for `ElementDescriptor` – it sits nicely alongside file descriptor from Unixland, font descriptor from UIKit, etc.

JPToroDev:
Excellent! Will get this updated and merged.

JPToroDev:
Moved to #563 .

PR #534, opened 2025-02-16
Remove required `page` initialization parameters from `Head` and `Body`.

JPToroDev:
This PR eliminates `PageContext`, creating individual environment properties for each of `PageContext.current`'s properties.

Now we can access a page's title anywhere like this: `@Environment(\.pageTitle) var pageTitle` —among other page metadata.

As a result of these changes:

1. `Layout.page` has transitioned from returning a `Page`, to returning `some HTML`. This means the property no longer requires a containing `HTML` type like `Section`.

2. `Head` and `Body` have dropped their required `page` initialization parameters.

3. For `Head`, dropping `page` in one initializer made it conflict with a sibling initializer. So, to retain the ability to use `Head` both with and without standard headers, I had to create `standardHeadersDisabled()` as a replacement.

4. Some tests were updated. More involved, others have been temporarily disabled and will be revisited in a follow-up PR. To reduce the size of this PR, I can undo the changes to any updated tests and just temporarily disable all of them.


twostraws:
Would it be sensible to have a single `\.page` key in the environment, rather than a bunch of separate properties?

JPToroDev:
The only issue with that is that `Layout` has an implicit `page` property already, which could cause a compiler conflict if someone tried to do `@Environment(\.page) var page`.

JPToroDev:
Perhaps we change that to `pageContent`? It's just the HTML body content of a page minus any page metadata.

JPToroDev:
Or perhaps the more sensible thing would be to make the environment property `\.pageMetadata` or the like.

JPToroDev:
Circling back here since this is a bottleneck for a couple other PRs.

My final recommendation would be to consolidate all page properties into `\.page` and rename `Layout.page` (which is just `some HTML` now) to either `contentArea` or `pageContent`.

PR #536, opened 2025-02-16
Add ability to style code blocks with line numbers, both at the element and site level.

JPToroDev:
This PR leverages Prism's official plugins to bring code-block line numbering and line highlighting to Ignite.

This entails adding a couple new JS methods to prism-core.js and some minor CSS in a new prism-plugins.css file.

At the site level:

`Site`'s current `syntaxHighlighters` property has been replaced with a new `syntaxHighlighterConfiguration` property, where you can configuration highlighter languages in addition to global line-number visibility and line wrapping behavior. This type will be a great home to any additional highlighter configuration settings `Site` may need in the future.

At the element level:

`CodeBlock` has received several new methods to highlight specific lines and line ranges, and to override the global line-number visibility settings.


twostraws:
Great idea!

PR #538, opened 2025-02-16
Eliminate problematic `BlockHTML` protocol.

JPToroDev:
Reopening of #463 .

This PR eliminates the problematic `BlockHTML` protocol, applying `HorizontalAligning` to applicable `HTML` types. It also renames private helper methods of `aspectRatio()` modifier that raise a compiler error post migration.

This fixes a huge mischaracterization within Ignite: when custom HTML elements are rendered, their content is wrapped in a `<div>`, which means they are inherently block-level elements. But Ignite wouldn't recognize them as such because they conform only to `HTML`. By treating `HTML` elements as block-level by default, we solve this issue while eliminating a bunch of boilerplate from the code base.

**NB**: After removing the `BlockHTML` variants of `aspectRatio()`, the compiler had an issue with a couple helper methods, also named `aspectRatio()`, exclusive to that file.

Both the public modifiers and these helper methods look exactly the same, so it might look like the public `aspectRatio()` has changed, but it has not. To make this clear, I've changed the helper methods to `applyAspectRatio()` and made the extension in which they're contained private.

This was the only compiler oddity that arose, and was necessary to address to make this build compile.

Also, there are several former `BlockHTML` types that I haven't conformed to `HorizontalAligning` because they didn't appear to have a valid need for it. Let me know if any types missing the conformance should have it.


twostraws:
This is much clearer now – thank you!

PR #540, opened 2025-02-17
Transition `MediaQuery` enum to `Query` protocol.

JPToroDev:
This PR transitions the `MediaQuery` enum to the `Query` protocol for greater:

### 1. Clarity

There are several cases in the `MediaQuery` enum that aren't actual CSS media queries, making the current name slightly misleading. `Query` creates a nice connection to media queries, while also broadening the type's scope.

### 2. Flexibility

Individual queries can more easily implement properties and methods that are relevant only to themselves.

### 3. Expandability

Users can now create media queries that Ignite has yet to implement.

This PR updates the types reliant on `MediaQuery`: `FontModifier`, `HiddenModifier`, and `CSSManager`.

This change has no breakage—the public API remains exactly the same—and updates all tests.

Lastly, this PR lays the groundwork for introducing the `Style` protocol, an API for generating custom CSS classes.


JPToroDev:
Thoughts on this? Again, no changes to the public-facing API. The PR just makes query types more robust.

twostraws:
Thank you! I'm going to merge this in, although I definitely do prefer names that are as specific as possible. I appreciate here that not all queries are media queries, so we'll proceed with the broader name.

JPToroDev:
Great! Might as well pull the all-nighter at this point. 😅

Follow-up PRs will be submitted shortly.

PR #541, opened 2025-02-17
Fix user-defined elements' ignoring custom attributes.

JPToroDev:
This PR migrates Ignite's 5 primary attribute methods—`style()`, `class()`, `data()`, `attribute()`, `id()`—to the `HTMLModifier` protocol so that custom attributes are recognized by custom elements.

This PR also fixes two errors that revealed themselves during the migration: extra spaces in the `GridTests` output strings, and a more reliable implementation of `ModifiedHTML`'s `init` and `render` methods.


twostraws:
Excellent!

PR #563, opened 2025-02-20
Rename `CoreAttributes` to `ElementDescriptor`.

JPToroDev:
Fresh PR for #531 . Take a look and see if this feels better than what we currently have.


JPToroDev:
Something about this code isn't sitting right with me. 🤔

Since this name doesn't affect the public API, I'm going to close this PR for the time being, and we can revisit it in the future.

PR #599, opened 2025-02-21
Add `\.page` property to `@Environment` to give elements access to page metadata.

JPToroDev:
Now distilled to its most important parts for easier review.

### DIFF
--- a/Sources/Ignite/Elements/Body.swift
+++ b/Sources/Ignite/Elements/Body.swift
@@ -25,8 +25,8 @@ public struct Body: DocumentElement {
     }
 
     public init() {
-        let page = PublishingContext.default.environment.page
-        self.items = flatUnwrap(page.body)
+        let pageContent = PublishingContext.default.environment.pageContent
+        self.items = flatUnwrap(pageContent)
     }
 
     /// Removes the Bootstrap `container` class from the body element.
--- a/Sources/Ignite/Framework/Environment/EnvironmentValues.swift
+++ b/Sources/Ignite/Framework/Environment/EnvironmentValues.swift
@@ -60,7 +60,10 @@ public struct EnvironmentValues {
     public let builtInIconsEnabled: BootstrapOptions
 
     /// The current page being rendered.
-    var page: Page = .empty
+    public internal(set) var page: PageMetadata
+
+    /// The content of the current page being rendered.
+    var pageContent: any HTML = EmptyHTML()
 
     /// The current piece of Markdown content being rendered.
     var article: Content = .empty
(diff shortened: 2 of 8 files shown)


JPToroDev:
Refocused reopening of  #534.

twostraws:
I want to find a way to get this in, but I'm concerned about it making things rather tangled. We have `Content` right now, which refers to a piece of Markdown, and we have `Page`, which refers to the metadata of a single page on the site. Names like `pageContent` seem likely to cause confusion.

Would something like `body` work?

JPToroDev:
`Page` has become `PageMetadata`, so hopefully it makes that type clear.

We can't use `body` because of the view's `body` var, but here are some alternatives:

`pageBody`
`documentBody`
`bodyArea`
`pageContainer`

But anything with `body` strikes me as confusing because of the existing property.

I have to say though, there are many times where the neutral term "content" would have be very handy to use throughout our recent work—associated type names, suffixes to variables like this, etc.

It does make me wonder if `Content` is too broad of a name for our Markdown content.

Since that type is only ever directly used in `ContentLayout.content`—`@Environment(\.content)` uses `ContentLoader`—it makes me wonder if we could do better with that name.

Changing it wouldn't break anything since it's never instantiated by users.

twostraws:
Ah! Yes, that makes a lot more sense; renaming `Content` would free up a significant chunk of the namespace. `Document` definitely doesn't work now, so maybe `Article`? That would leave `content` for the HTML of a layout, and `page` for the page metadata.

JPToroDev:
I like `Article`!

We'd also need to update `ContentLayout` to match – which ties into our other big point of type name confusion.

Since the `content` property of `Layout` references `ContentLayout.body`, perhaps we could rename `ContentLayout` to `ArticleContent`?

That feels little vague though...

In the past I had suggest the suffix -Template, but you had mentioned wanting to reserve that namespace should we ever integrate an existing templating engine in the future. Does that feeling still hold true?

JPToroDev:
I've thought about alternate names for `Content` for quite a bit, and pickings are slim.

The only other one that might be worth considering is `Matter`. Although on the surface it seems vague, one of its definitions communicates exactly what it is:

"the substance or content of a text as distinct from its manner or form"

That could free up `Article` to use elsewhere, and create a nice relationship between the two types.

twostraws:
Given that the opening comment for `ContentLayout` is "Content layouts describe custom layout structures for articles", I think the decision here is straightforward: let's move to Article rather than Content.

Getting a templating engine in place remains a high-priority task – something like <https://github.com/hummingbird-project/swift-mustache> would be an excellent choice, and make a lot of people happy.

JPToroDev:
Yes, `Article` is unequivocally the best name.

Re our -`Layout` types: they represent the page-body content of a `Layout` (full-page layout including navigation bar and footer). 

In fact, these -`Layout` types populate `Layout`'s `content` property. -`Layout`s and `Layout`were at one point independent, but that is no longer the case.

So rather than rename `ContentLayout` to `ArticleLayout`, what better scheme could we come up with to squash this source of confusion once and for all?

The former names of `StaticPage` and `TagPage` work fine. `ArticlePage` feels a little odd, but maybe not. 

The most accurate names would use the `Content` namespace. Do `PageContent`, `ArticleContent`, and `TagContent` (or maybe `TaggedContent`?) feel right? 

I still think a broader term for the last—like `CategoryContent` or `ArchiveContent`—would be the most flexible/stable long term.

PR #600, opened 2025-02-21
Consolidate “site” prefixed environment variables into one `\.site` property, mirroring `\.page`.

JPToroDev:
To create symmetry and consistency with #599.


twostraws:
Great! Thank you!

PR #603, opened 2025-02-22
Add `Category` protocol to create foundation for categorizing articles beyond tags.

JPToroDev:
In addition to grouping articles by tag, many blogs also group articles by date and author.

While this PR does not introduce that functionality, it lays the groundwork for doing so, ensuring that the current API can evolve without creating breaking changes.

This PR introduces the `Category` protocol, which provides `TagLayout` with more detailed information about the category it is receiving. 

Currently, `TagLayout` will receive one of two concrete implementations: `TagCategory` or `AllTagsCategory`. 

But in the future `TagLayout` could receive `DateCategory`, `AuthorCategory`, etc.

This feature ties into a larger conversation around updating `TagLayout` with a clearer and more comprehensive name.

PR #640, opened 2025-02-25
Improve names of -`Layout` types.

JPToroDev:
Based on our conversations, I’ve circled on some promising names for our layout types:

#### `StaticLayout` to `Page`
- Unlike the other two types, `Page`s are not reusable layouts; they are individual pages.
- `StaticLayout.parentLayout` becomes `Page.layout`. Very clear.

#### `ArticleLayout`
- Type name remains the same.
- `ArticleLayout.parentLayout` become `ArticleLayout.pageLayout` to create clear relationship between layouts.

The one that might require further discussion:

#### `TagLayout` to `ArchiveLayout`
- Broader name with long-term stability #603
- Lacks ambiguity of `TagLayout`—this isn't the layout of a tag.
- Creates a namespace distinct from `Tag`, the `HTML` type.
- `TagLayout.parentLayout` become `ArchiveLayout.pageLayout` to create clear relationship between layouts.

### DIFF
--- a/Sources/Ignite/Elements/Link.swift
+++ b/Sources/Ignite/Elements/Link.swift
@@ -92,7 +92,7 @@ public struct Link: InlineElement, NavigationItem, DropdownItem {
     /// - Parameters:
     ///  - target: The new target to apply.
     ///  - content: The user-facing content to show inside the `Link`.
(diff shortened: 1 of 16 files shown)


twostraws:
I've stared at this for days, and still don't understand the rationale behind these renaming. I'm going to close the issue; I think we need to set up a FaceTime call to discuss this further.

PR #658, opened 2025-03-02
Set HTML lang attribute based on Site language.

mwermeester:
Hello,

By default, the HTML document language defaults to English:

`<html lang="en">`

The `Site` struct allows setting a language via the `language` variable, however this is never taken into account and the language always defaults to English. 

This pull requests provides a way to take into account the `Site` language and to set it as the language of the HTML document.

I understand that there is a more extensive language management underway, so consider this as a temporary way to address the issue. 

Thanks!

### DIFF
--- a/Sources/Ignite/Elements/Document.swift
+++ b/Sources/Ignite/Elements/Document.swift
@@ -25,7 +25,7 @@ struct Document: HTML {
 
     func render() -> String {
         var attributes = attributes
-        attributes.add(customAttributes: .init(name: "lang", value: language.rawValue))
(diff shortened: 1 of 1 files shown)


JPToroDev:
Fantastic—thanks! If you fix the test failures I'll get his merged in ASAP.

mwermeester:
Thank you, I fixed the test.

mwermeester:
Simplified the code a bit by removing the language parameter as the language will be initialized based on what's in the PublishingContext.

JPToroDev:
Stellar—thank you!

PR #698, opened 2025-03-10
Add convenience property to disable auto-margins on `VStack` / `ZStack` children.

JPToroDev:
This turned out to be a huge pain point while migrating my sites—particularly for `VStack`, which offers spacing of its own. I think this solution provides a convenient fix while keeping the default behavior as expected.

### DIFF
--- a/Sources/Ignite/Elements/VStack.swift
+++ b/Sources/Ignite/Elements/VStack.swift
@@ -22,6 +22,10 @@ public struct VStack: HTML {
     /// The child elements contained in the stack.
     private var content: HTMLCollection
 
+    /// Whether the `VStack` should ignore any margins automatically
+    /// applied to its children, like the bottom margin of a paragraph.
+    private var shouldIgnoreMargins: Bool = false
+
     /// Creates a new `Section` object using a block element builder
     /// that returns an array of items to use in this section.
     /// - Parameter items: The items to use in this section.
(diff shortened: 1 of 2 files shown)


twostraws:
This is an important feature, but I think we should clarify the external naming here, because "normalize" isn't really that clear. Are we saying that with this setting views go edge to edge?

Alternatively: the reason we're looking at this is because it caused a difference in the try! Swift Tokyo website code, so perhaps the real solution here is to stick with the previous "always remove the margins" behavior, and just tell folks it's a feature – have them adding the margin back by hand, if it's something they want.

Thoughts?

JPToroDev:
The reason there was a difference in the try! Swift Tokyo website wasn't actually because the `ZStack` / `VStack` behavior changed—those types have removed bottom margins from the start, which is important, because unlike a regular `Section`, these types are meant to be positioning contexts.

The discrepancy in the try! Swift Tokyo website was caused because we went from ordering attributes by collection order, to alphabetic order. This change is actually a problem, and I'm working on a PR to revert to the previous behavior now.

All to say reverting to the previous implementation of removing margins by default would be a much simpler, more appropriate fix.

twostraws:
> All to say reverting to the previous implementation of removing margins by default would be a much simpler, more appropriate fix.

Okay, let's do that. We should add documentation saying this is intended behavior, to ensure things like `.bottom` actually means the bottom.

JPToroDev:
You got it.

PR #703, opened 2025-03-10
Remove hardcoded theme value in `NavigationBar`. Remove theme attribute from `Document`.

JPToroDev:
`navigationBarStyle(.dark)` will override:
 
1. The drop-down button text color
2. The background of the drop-down and the the text color of each drop-down item

So in IgniteSamples, the above modifier will not only make the drop-down button text white—which looks good against a red background—but it will also make the drop-down modal off-black in both dark and light modes.

Is that the behavior we want? If not, should we update the above modifier to be something like this?

```swift
navigationBarStyle(_ style: NavigationBarStyle, for component: NavigationBarComponent)
```

Where `NavigationComponent` has cases like `.button`, `menu`, `.all`?

### DIFF
--- a/Sources/Ignite/Elements/Document.swift
+++ b/Sources/Ignite/Elements/Document.swift
@@ -26,7 +26,6 @@ struct Document: HTML {
     func render() -> String {
         var attributes = attributes
         attributes.add(customAttributes: .init(name: "lang", value: language.rawValue))
-        attributes.add(customAttributes: .init(name: "data-bs-theme", value: "auto"))
         var output = "<!doctype html>"
         output += "<html \(attributes)>"
         output += contents.map { $0.render() }.joined()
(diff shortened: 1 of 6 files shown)


JPToroDev:
Closing this and implementing a much simpler fix—an extra `init` to `Dropdown` that allows us to override the text color of dropdown titles:

```swift
    public init(
        @ElementBuilder<any DropdownItem> items: () -> [any DropdownItem],
        @InlineElementBuilder title: () -> any InlineElement
    ) {
        self.items = items()
        self.title = title()
    }
```

This, coupled with removing the hardcoded color scheme in `NavigationBar`, gets things working as expected.

JPToroDev:
I've worked out a much simpler solution: Adding a new `init` to `Dropdown` that allows users to easily style their title, which is `InlineElement`.

Not many modifiers work with `NavigationItem` at the moment, and adding support by extending `NavigationItem` doesn't create the expected behavior—at least for `foregroundStyle()` and `Dropdown`.

The new `init` is a good one to have regardless, and this is a simple, dynamic fix.

Issue #735, opened 2025-03-18
Struggling positioning elements horizontally, one after the other


cepages:
I'm trying to have a Dropdown leading and then a Text centered

I've tried multiple things, included the advised in [here](https://github.com/twostraws/Ignite/issues/6). The response says to use a section:
```
        Section {
            Dropdown("Menu") {
                Link("Experience", target: Experience())
                Link("Hobbies", target: Hobbies())
                Divider()
                Text("Change language")
                Link("Spanish", target: "/")
            }
            .role(.primary)
            Text("My site").horizontalAlignment(.center)
                .border(.red)
        }
```
![Image](https://github.com/user-attachments/assets/d2b83ace-d347-4ca8-b22b-bf1921b78084)

But the link actually uses a `Text`(https://github.com/twostraws/IgniteSamples/blob/main/Sources/Components/SocialFooter.swift) :

```
      Text {
            Dropdown("Menu") {
                Link("Experience", target: Experience())
                Link("Hobbies", target: Hobbies())
                Divider()
                Text("Change language")
                Link("Spanish", target: "/")
            }
            .role(.primary)
            "Hello"
            "Bye"
            "Nooo"
        }
        .horizontalAlignment(.center)
        .margin(.top, .xLarge)
        .font(.title2)
```

![Image](https://github.com/user-attachments/assets/3aa9975e-532f-44d7-a39c-1129dfeefd78)

The only way I could achieve having them together is with a Grid. The drop down will be leading, but because each column will have the same space, the text won't be just after the dropdown.

JPToroDev:
Funny timing—I was attempting the same layout yesterday. I'll make a PR with my solution, as Ignite should make this easy to do out of the box.

cepages:
Haha yes funny timing, although are we talking about the same thing? The PR covers a new element which places a text and an image side by side.

My issue in particular was using a dropDown and a text side by side, although probably the best solution should allow any `InlineElement`, something like a HStack.

Maybe the problem is that in HTML is that straightforward and `flex` would be needed? if so is there a hack to achieve it?

I guess, creating a new element that uses the render output of N `InlineElements` between spans would not work?:

```
public func render() -> String {

     return "<span\(attributes)>\(inlineElement1)...\(inlineElementN)</span>"
}
```

JPToroDev:
I agree that `HStack` would be a nice addition—it's been on the todo list for a while.

[...]

PR #739, opened 2025-03-19
Create `Label` type to easily place `Text` and `Image` side by side.

JPToroDev:
This PR addresses an issue raised in #735 by creating a SwiftUI-like `Label` type that easily allows users to put `Text` and `Image` side by side—particularly convenient for nav bars.

### DIFF
--- a/Sources/Ignite/Elements/Label.swift
+++ b/Sources/Ignite/Elements/Label.swift
@@ -1,0 +1,71 @@
+//
+// Label.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+/// A view that displays an icon and text side by side, with a default
+/// spacing of 10 pixels in-between. To adjust the spacing, use `margin()`
+/// on either `title` or `icon`.
+public struct Label: InlineElement {
+    /// The content and behavior of this HTML.
+    public var body: some HTML { self }
+
+    /// The standard set of control attributes for HTML elements.
+    public var attributes = CoreAttributes()
+
+    /// Whether this HTML belongs to the framework.
+    public var isPrimitive: Bool { true }
+
+    /// The text content to display alongside the icon.
+    private var title: any InlineElement
+
+    /// The icon element to display before the title.
+    private var icon: any InlineElement
+
+    /// Creates a label with a string title and image icon.
+    /// - Parameters:
+    ///   - title: The text to display in the label.
+    ///   - image: The image to use as the label's icon.
+    public init(_ title: String, image: Image) {
+        self.title = title
+        self.icon = image
+    }
+
+    /// Creates a label with custom title and icon content.
+    /// - Parameters:
+    ///   - title: A closure that returns the label's text content.
+    ///   - icon: A closure that returns the label's icon content.
+    public init(
+        @InlineElementBuilder title: () -> some InlineElement,
+        @InlineElementBuilder icon: () -> some InlineElement
+    ) {
+        self.title = title()
+        self.icon = icon()
+    }
+
+    public func render() -> String {
+        var title = title
+
+        if !(title is String) {
+            title.attributes.add(classes: "mb-0")
+        }
+
+        var icon = icon
+
+        if !(icon is String) {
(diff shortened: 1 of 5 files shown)


Sources/Ignite/Elements/Label.swift
```diff
@@ -0,0 +1,71 @@
...
+    private var title: any InlineElement
+
+    /// The icon element to display before the title.
+    private var icon: any InlineElement
+
+    /// Creates a label with a string title and image icon.
+    /// - Parameters:
+    ///   - title: The text to display in the label.
+    ///   - image: The image to use as the label's icon.
+    public init(_ title: String, image: Image) {
```
twostraws, line 31: If we're to have an `image` overload, I think it ought to be a string to match what SwiftUI does.

Sources/Ignite/Elements/Label.swift
```diff
@@ -0,0 +1,71 @@
...
+        self.icon = image
+    }
+
+    /// Creates a label with custom title and icon content.
+    /// - Parameters:
+    ///   - title: A closure that returns the label's text content.
+    ///   - icon: A closure that returns the label's icon content.
+    public init(
+        @InlineElementBuilder title: () -> some InlineElement,
+        @InlineElementBuilder icon: () -> some InlineElement
```
twostraws, line 42: Folks who want a precise Image – a specific size, for example – should use this overload. We might also want a systemImage overload for icons.

cepages:
I found the label works fine except for how much rooms it takes. 

In the following example the `Text` takes all the width but the `Label` just adjust to its content.

Image

Unless this is what it supposes to achieve, can this be fixed if we use `<div>` instead of using `<span>`?

or make it block-level:

```
span {
    display: block;
}
```

Sources/Ignite/Elements/Label.swift
```diff
@@ -0,0 +1,71 @@
...
+
+        if !icon.attributes.styles.contains(where: { $0.property == Property.marginRight() }) &&
+           !title.attributes.styles.contains(where: { $0.property == Property.marginLeft() }) {
+            icon.attributes.append(styles: .init(.marginRight, value: "10px"))
+        }
+
+        var attributes = attributes
+        attributes.append(styles: .init(.display, value: "inline-flex"))
+        attributes.append(styles: .init(.alignItems, value: "center"))
+        return "<span\(attributes)>\(icon)\(title)</span>"
```
cepages, line 69: I found the label works fine except for how much rooms it takes. 

In the following example the `Text` takes all the width but the `Label` just adjust to its content.

Image

Unless this is what it supposes to achieve, can this be fixed if we use `<div>` instead of using `<span>`?

or make it block-level:

```
span {
    display: block;
}
```

JPToroDev: @cepages The current behavior is intended. I think `HStack` will achieve what you're looking for. I'll try to submit that soon.

Sources/Ignite/Elements/Label.swift
JPToroDev: That's where my mind first went too, but that'd never allow users to add a description to their image. 

Maybe we have `Label` just use decorative images for this initializer?

Sources/Ignite/Elements/Label.swift
JPToroDev: I've pushed a commit with the revised initializers. Folks who want to add an image description can use this init.

Sources/Ignite/Elements/Label.swift
twostraws: In SwiftUI, the text *is* the description for the image. Are you anticipating that being different here? If so, we need to be careful around naming the thing.

JPToroDev: Brain fart on my end. Yes, the text should be the description here, too.

JPToroDev: Should be good to go now.

Sources/Ignite/Elements/Label.swift
cepages: Fab, thanks for the confirmation.

PR #741, opened 2025-03-19
Allow users to determine the style of an article's tag links.

JPToroDev:
This PR allows users to determine the appearance of an article's tag links by updating the `tagLinks()` method from this:

```swift
 @InlineElementBuilder public func tagLinks() -> (some InlineElement)? {
```

to this:

```swift
  public func tagLinks(style: TagLinkStyle = .automatic) -> [Link]? {
```


twostraws:
Love it – thank you!

PR #751, opened 2025-03-20
`Frame`: Restore default alignment to `.center` and fix longstanding bugs.

JPToroDev:
`Frame` has never been able to handle cases like this correctly:

```swift
Image("/images/interactions.png")
    .resizable()
    .frame(height: .px(250))
    .frame(width: .percent(100%), height: .px(550), alignment: .bottom)
```

This PR has reimplemeneted `Frame` to fix that issue.

Internally throughout Ignite, when we need to contain an element in a `<div>` wrapper, we do so by enclosing the view in a `Section`.

But for very temperamental cases like `<div>`-based positioning contexts, we need to be able to determine whether a `<div>` was created:

1. Explicitly by the user as a structural element.
2. Implicitly by the framework, in which case the type of its contained content might be important to check.

Here I've proposed a new internal type called `Container` to serve this purpose. 

When we encounter a `Container`, we know the identity of its contained content is important, and therefore know to query it.

There are a few band-aids throughout the framework that this draft highlights—like how we determine when to wrap a view in a `Container` in `foregroundStyle()`—that will need to be addressed in separate PRs.

Once we iron out `Container`, I'll move that to its own PR as well.

### DIFF
--- a/Sources/Ignite/Elements/Grid.swift
+++ b/Sources/Ignite/Elements/Grid.swift
@@ -106,7 +106,7 @@ public struct Grid: HTML, HorizontalAligning {
         case .exact(let pixels) where pixels != 0:
             gridAttributes.append(styles: .init(.rowGap, value: "\(pixels)px"))
         case .semantic(let amount) where spacingAmount != .semantic(.none):
-            gutterClass = "gy-\(amount.rawValue)"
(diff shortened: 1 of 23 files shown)


JPToroDev:
Will open a new PR with a better approach.

PR #755, opened 2025-03-22
Implement `HStack`.

JPToroDev:


Tests/IgniteTesting/Elements/HStack.swift
```diff
@@ -0,0 +1,48 @@
...
+
+/// Tests for the `HStack` element.
+@Suite("HStack Tests")
+@MainActor
+class HStackTests: IgniteTestSuite {
+    @Test("HStack with elements")
+    func basicHStack() async throws {
+        let element = HStack(alignment: .top) {
+            FormFieldLabel(text: "Top Label")
+            FormFieldLabel(text: "Bottom Label")
```
MrSkwiggs, line 21: Top & Bottom nomenclature here might make things confusing? 😬

twostraws:
We need to be very careful with the nomenclature here – I don't really want `HStack` to be the first port of call for folks, because a true HStack ought to keep its contents horizontal even when space is constrained, whereas a mobile-friendly website should adapt based on snapping points. That doesn't mean `HStack` is unnecessary, I just want to make sure folks are choosing it for the right reasons, and understand what they will get.

JPToroDev:
Agreed. Since our `HStack` mirrors the Bootstrap one, I'm hopeful folks won't find the nomenclature too unclear.

PR #761, opened 2025-03-22
Allow multiple sites to run at once.

JPToroDev:
This PR allows multiple Ignite sites to run at once. Ignite runs the first site on port 8000 and increments the port number by 1 for every subsequent site.


twostraws:
Nice – thank you!

PR #762, opened 2025-03-24
Fix incorrect access level for class modifier

MrSkwiggs:
The variadic `class` modifiers were not marked as public but they are very handy, so I marked them as public.


twostraws:
Thank you!

PR #765, opened 2025-03-26
Make `EnvironmentValues` available in `prepare()`.

JPToroDev:
This PR gives users access to a subset of available environment values—like a site's articles—in `Site`'s `prepare()` method.

### DIFF
--- a/Sources/Ignite/Framework/Environment/DynamicEnvironmentValues.swift
+++ b/Sources/Ignite/Framework/Environment/DynamicEnvironmentValues.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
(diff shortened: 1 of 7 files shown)


twostraws:
I'm keen to avoid leaking implementation details if possible; if we can get @Environment to work in Site it would be preferable to passing in values. For example, in the case of making content available in prepare(), could we push the content parsing phase up so that it happens before prepare() is called?

JPToroDev:
I also took a bit of inspiration from `SwiftUI`'s `ShapeStyle`.

JPToroDev:
That makes sense. I'll submit a new PR with that approach. Then I think we're good to push a small tag and set our sights on the multilingual stuff.

PR #774, opened 2025-03-27
Fix positioning issues caused by nested `<div>`s in `Grid`.

JPToroDev:


Sources/Ignite/Elements/Grid.swift
```diff
@@ -149,6 +149,7 @@ public struct Grid: HTML, HorizontalAligning {
                 } else {
                     handleItem(item)
                         .class(gutterClass)
+                        .render()
```
MrSkwiggs, line 152: (for my understanding) why is calling `render()` here needed (but not in the other branches of this control flow)?

MrSkwiggs (approved):

twostraws:
Thank you!

Issue #752, opened 2025-03-20
Small fontStyle


mwermeester:
For fontStyle, we have the options: body, lead, title1, title2, etc.)

I would like to have a "small" fontStyle. 
e.g., for Footer elements.

PR #796, opened 2025-04-03
Add small font styles.

JPToroDev:
This PR addresses #752, solving a common pain point when sizing text.

This PR uses the pattern `footer1`, `footer2`, etc.—mirroring`titleN` and avoiding the confusion that accompanies a naming system like `subheadline`, `footer`, `caption`.

### DIFF
--- a/Sources/Ignite/Elements/Text.swift
+++ b/Sources/Ignite/Elements/Text.swift
@@ -113,7 +113,7 @@ public struct Text: HTML, DropdownItem {
 extension HTML {
     func fontStyle(_ font: Font.Style) -> any HTML {
         var copy: any HTML = self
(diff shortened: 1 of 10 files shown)


MrSkwiggs (approved):

twostraws:
This is a great feature, thank you! I think we should look for a better name, though – `footer` is something that comes at the end of a page, rather than a small size. Would `small` then variants of `xSmall` work? Certainly we could get down to `xxxSmall` without it being strange. We're kind of stuck with `title3` etc because of the big HTML/Bootstrap precedent.

JPToroDev:
Love it!

JPToroDev:
Updated in v0.6-beta branch.

PR #805, opened 2025-04-16
Fix for Article.type always returning an empty string.

gavinjerman:
Currently the `Article type` variable always returns an empty string. It uses the following code to extract the `type` string from the `Article`'s metadata:

```
if let type = metadata["type"] {
            type as? String ?? ""
        } else {
            fatalError("""
            Unable to retrieve type for article '\(title)'. \
            Please file a bug report on the Ignite project.
            """)
        }

```

However, the cast `type as? String ?? ""` will always fail and return the default empty string. The `metadata["type"]` is stored in `Article`'s `init()` using the following code:

```
// Save the first subfolder in the path as the article's type
        let pathParts = path.split(separator: "/") // removes empty
        if pathParts.count > 1 { // no type if not in subdirectory
            metadata["type"] = pathParts[0]
        }
```

The problem here is `path.split(separator: "/")` returns a `[Substring]` not a `[String]`, so `type` is stored as a `Substring` not a `String`, meaning the cast `type as? String ?? ""` will always return an empty string.

A side effect of this is the `ArticleLoader` func `typed(_ type: String)` will always return an empty `[Article]`. This is how I noticed the bug :)

The fix takes the result of `path.split(separator: "/")` and maps it to a `[String]` to ensure the `metadata["type"]` is stored as a `String`.

### DIFF
--- a/Sources/Ignite/Framework/Article.swift
+++ b/Sources/Ignite/Framework/Article.swift
@@ -158,7 +158,7 @@ public struct Article {
         self.path = metadata["path"] as? String ?? deployPath
 
         // Save the first subfolder in the path as the article's type
-        let pathParts = path.split(separator: "/") // removes empty
+        let pathParts = path.split(separator: "/").map { String($0) } // removes empty
         if pathParts.count > 1 { // no type if not in subdirectory
             metadata["type"] = pathParts[0]
         }


gavinjerman:
Hmm, to my shame, I don't know how to do tests. I missed Paul's event on that. Is there any documentation, notes, etc. on adding tests to Ignite?

JPToroDev:
Nothing to be ashamed about! Many recent test makers had no prior experience.

This would be a pretty sophisticated first test to write, so I'm happy to merge this in and provide the test. 

But to ignite your test-writing journey, I'd love to see you contribute a test that's smaller and more contained!

There are no formal instructions about adding tests to Ignite, but if you look at existing tests, starting with the simplest ones first, I think you'll very quickly understand how they work.

PR #806, opened 2025-04-16
Perform Document rendering in reversed order to defer Head rendering

TomaszLizer:
While tinkering with Ignite I noticed one bug with regards to CodeBlock rendering. It is supposed to automatically pick up Code language and highlight it when using Ignition DSL. This was not working though for the first page with CodeBlock element.

After some debugging I have noticed that we update `PublishingContext` with language of the code block and based on that correct header is to be rendered. Unfortunatelly at that time `Head` of that page is already rendered, hence this update to the context is not reflected.

I have considered different approaches:
- Leveraging that we expect `Document` to have `Head` element - getting its index (though I believe it should always be 0?) and deferring its rendering phase "manually"
  - More explicit, yet too much hassle with no real gain IMO
- Reversing render phase, then reversing again to construct content
  - ✅ Chosen solution - because of its simplicity
- Spliting render into prep and render phase (eg add pre-render phase)
  - Might be nicest solutjon, yet super complex and do not see a reason for such big refactor (targeting single case)
- Remove automatic code highlight insertion into `PublishingContext`
  - Maybe it should just never be there and users should be forced to correctly reference languages for code highlight in `Site` definition?

Attaching video with issue reproduced:

https://github.com/user-attachments/assets/832c5a24-5201-4a6a-ba41-3bb8487bb4da

Issue can be reproduced by having site with homePage having such body:
```swift
var body: some HTML {
    Text("Hello world!")
        .font(.title1)
    ForEach(articles.all) { article in
        Link(article.title, target: article)
    }
    
    CodeBlock(.swift) {
        """
        struct Test {
            var test = "Hello World"
        }
        """
    }
}
```

Alternatively it can be reproduced with https://github.com/twostraws/IgniteSamples when commenting out `syntaxHighlighterConfiguration` in `Site` definition.

### DIFF
--- a/Sources/Ignite/Elements/Document.swift
+++ b/Sources/Ignite/Elements/Document.swift
@@ -27,8 +27,11 @@ public struct Document: MarkupElement {
         attributes.append(customAttributes: .init(name: "lang", value: language.rawValue))
         var output = "<!doctype html>"
         output += "<html\(attributes)>"
-        output += head.markupString()
-        output += body.markupString()
+        let bodyMarkup = body.markup()
+        // Deferred head rendering to accommodate for context updates during body rendering
+        let headMarkup = head.markup()
+        output += headMarkup.string
+        output += bodyMarkup.string
         output += "</html>"
         return Markup(output)
     }


MrSkwiggs:
Thank you for the contribution @TomaszLizer. 
While your solution works out of the box, I do not believe it is the best way of handling this particular scenario. 

I do like suggestion 3 however, but as you mentioned this would incur a significant increase in scope for this PR.

I'll convene with the team to see exactly what direction we want to take, and if you're still up for it by then, we will definitely appreciate your contributions.

We'll keep you posted right here!

cc. @JPToroDev

TomaszLizer:
Adding additional handling in `@DocumentBuilder` seems fine to me. While investigating I stumbled on one potential issue.
In theory someone could provide multiple headers which should be not supported. Wanted to add precondition failure for such - yet `Never` is not supported by result builders 🤦 

For now I have came up with split of components using reduce and using last `Head` provided.

```swift
    public static func buildBlock(_ components: any DocumentElement...) -> some HTML {
        Document {
            let (header, components) = components.reduce(
                into: (header: Head, components: [any DocumentElement])(Head(), [])
            ) { partialResult, element in
                if let header = element as? Head {
                    // In case multiple headers were provided only last one will be used
                    partialResult.header = header
                } else {
                    partialResult.components.append(element)
                }
            }
            
            header

            // Add all provided components
            for component in components {
                component
            }
        }
    }
```
WDYT @MrSkwiggs @JPToroDev ?
Or maybe there is some way to crash resultBuilder in runtime or validate in buildtime? :)

JPToroDev:
Excellent catch!

We'd also need to check for multiple `Body` elements, too.

I like your approach of using the last one, but we'd need to add a build warning like:

`PublishingContext.shared.addWarning("Multiple Head elements detected, only the last will be used.")`

Otherwise, we could throw a fatal error, but that seems a little extreme.

JPToroDev:
Rather than a warning, perhaps making this an official `PublishingError` makes the most sense.

TomaszLizer:
I am back after too many rounds of [sałatka jarzynowa](https://pl.wikipedia.org/wiki/Sałatka_jarzynowa) 😅 

In the meantime I was considering different solutions to the above problem and I only found more doubts.
As of current implementation we assure that there is a `Head` component by adding default one if none is provided. We do 't do the same for `Body`. After discussions we had this PR currently alters `body` that user creates

JPToroDev:
You continue to do truly exceptional work—thank you!

You're absolutely right: currently folks can conform types to `DocumentElement`, but that should not be the case. 

In fact, since `Body` and `Head` are the only valid children of a `Document` (`<html>`), we should do away with the `DocumentElement` protocol altogether.

I've done that in the [v0.6-beta branch](https://github.com/twostraws/Ignite/tree/v0.6-beta).

Take a look at that branch, and let me know if it gives you a clear direction from what you've outlined above.

TomaszLizer:
What you have in the beta branch is basically how I imageined it to be best case scenario- just hold body and head directly. Then rendering becomes trivial, we just do that for body first.
Would you like then to include this as part of your beta branch changes (head rendering defer) @JPToroDev?

JPToroDev:
Excellent!

> Would you like then to include this as part of your beta branch changes...

That'd be great!

TomaszLizer:
@JPToroDev Rebased on beta branch and updated to accomodate for latest changes.

JPToroDev:
Beautiful! Thanks for the great work on this 🙌🏼

PR #810, opened 2025-04-22
Add try Swift Tokyo in RADME.md

KaitoMuraoka:
Just as the title says.
I think this is a good example of managing an app and Ignite in a single repository.


twostraws:
Thank you!

PR #811, opened 2025-04-27
Fix custom font inline style

TomaszLizer:
While working with 0.6 beta I have noticed that custom fonts are not applying. It seems to be caused by redundant quotation  marks.

Here is how change manifests in html:
Screenshot 2025-04-27 at 21 22 35

After the fix fonts are correctly applied.


JPToroDev:
Thanks for flagging!

While it's valid for single-word fonts to lack quotes, multi-word fonts need them.

I've switched over to single quotes in the latest commit to fix the escaping issue.

Does that fix things?

TomaszLizer:
@JPToroDev With update to single quotes it now renders correctly in tested case :) 
Screenshot 2025-04-29 at 07 38 35

PR #814, opened 2025-04-29
Fix Missing Publishing Context HTTP Error

MrSkwiggs:
We actually weren’t passing the relevant HTTP errors to `ErrorPage`s (either we never did, or it broke at some point without anyone noticing). The existing tests only checked for static strings, so they didn’t catch that the actual error instances weren’t being passed around correctly.

That’s now fixed, and the tests have been updated to properly check for this going forward.

### DIFF
--- a/Sources/Ignite/Framework/Environment/EnvironmentValues.swift
+++ b/Sources/Ignite/Framework/Environment/EnvironmentValues.swift
@@ -105,7 +105,8 @@ public struct EnvironmentValues {
         site: any Site,
         allContent: [Article],
         pageMetadata: PageMetadata,
-        pageContent: any LayoutContent
+        pageContent: any LayoutContent,
+        httpError: HTTPError = EmptyHTTPError()
     ) {
         self.decode = DecodeAction(sourceDirectory: sourceDirectory)
         self.articles = ArticleLoader(content: allContent)
@@ -117,6 +118,7 @@ public struct EnvironmentValues {
         self.builtInIconsEnabled = site.builtInIconsEnabled
         self.timeZone = site.timeZone
         self.page = pageMetadata
+        self.httpError = httpError
 
         self.site = SiteMetadata(
             name: site.name,
--- a/Tests/IgniteTesting/Publishing/Site.swift
+++ b/Tests/IgniteTesting/Publishing/Site.swift
@@ -122,7 +122,11 @@ struct SiteTests {
 
     @Test("Site published with a custom ErrorPage and custom content")
     func publishingWithCustomErrorPageAndContent() async throws {
-        let errorPage = TestErrorPage(title: "A different title", description: "A different description")
+        let expectedError = PageNotFoundError()
+
+        let errorPage = TestErrorPage(title: "A different title", description: "A different description") { error in
+            #expect(error.statusCode == expectedError.statusCode)
+        }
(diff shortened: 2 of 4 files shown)


Tests/IgniteTesting/TestWebsitePackage/Sources/TestErrorPage.swift
```diff
@@ -13,7 +13,30 @@ struct TestErrorPage: ErrorPage {
...
+    }
+
     var body: some HTML {
+        ErrorChecker { errorChecker(error) }
         EmptyHTML()
     }
+
+    struct ErrorChecker: HTML {
+        init(handler: @escaping () -> Void) {
+            handler()
```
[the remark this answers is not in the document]
MrSkwiggs: This is simply a Testing utility; it only allows us to hook into & check the injected env. error during testing

Sources/Ignite/Framework/Environment/EnvironmentValues.swift
```diff
@@ -105,7 +105,8 @@ public struct EnvironmentValues {
         site: any Site,
         allContent: [Article],
         pageMetadata: PageMetadata,
-        pageContent: any LayoutContent
+        pageContent: any LayoutContent,
```
JPToroDev, line 108: Since `pageContent` isn't relevant here, can we create a dedicated init that replaces `pageContent` with `httpError`, instead of just adding it?

Tests/IgniteTesting/Publishing/Site.swift
```diff
@@ -93,7 +93,11 @@ struct SiteTests {
 
     @Test("Site published with a custom ErrorPage and custom content")
     func publishingWithCustomErrorPageAndContent() async throws {
-        let errorPage = TestErrorPage(title: "A different title", description: "A different description")
+        let expectedError = PageNotFoundError()
+
+        let errorPage = TestErrorPage(title: "A different title", description: "A different description") { error in
+            #expect(error.statusCode == expectedError.statusCode)
```
JPToroDev, line 99: Could we simply check `#expect(errorPage.error.statusCode == expectedError.statusCode)` instead of introducing the `errorChecker` closure?

JPToroDev (commented):
Great catch! Couple small notes, otherwise looks great!

Tests/IgniteTesting/Publishing/Site.swift
MrSkwiggs: Heya sorry for the delay, life happened. 

Unless I'm missing something, checking the page's error after calling `site.publish()` will not work as it's set (and reset) dynamically during publishing? 

What I want the test to do is to guarantee that the correct error is present _during_ site publishing. I can only do so by injecting this check when the rendering actually happens.

Sources/Ignite/Framework/Environment/EnvironmentValues.swift
MrSkwiggs: Not sure what you mean by this? `EnvironmentValues` has 5 different inits and I only modified the one that required the fix

Tests/IgniteTesting/Publishing/Site.swift
JPToroDev: Gotcha—thanks for clarifying!

JPToroDev (approved):

PR #815, opened 2025-05-01
Add support for hosting subsite locally

tockrock:
Hosts the generated site in a subdirectory locally
to keeping the same path for the assets.

This is achieved by the following methods
1. Removes the subdirectory from the accessed path (translate_path)
2. Transfer to the subdirectory when accessing root (do_GET)

~~The subsite is detected from the /index.html and where the css is specified.
This method was chosen since it was the same way a subsite was blocked to run on the server.~~

The subsite is now detected using the canonical url of the /index.html of the site for more robustness.


JPToroDev:
Thanks for submitting this great feature!

I know it's hard to write formal tests for something like this—are there any potential side effects or breakages that this change could introduce?

tockrock:
Thanks for having a look!

I haven't found a tests for the IgniteCLI code for both the Swift and Python code so far.
Let me know if I've missed them.

Assuming they don't exist, the best way forward would be to create one, but that would be a bit of work that probably would involve restructuring some of the code to be more testable.

As for potential issue, the Python code is fairly straight forward that I don't think it would cause issue on that side.

On the Swift side, I've changed the way to detect if it's a subsite by looking at the canonical url.
The two issue matching with "/css" was that a subsite can have multiple "/css" within it's path,
and the site's `useDefaultBootstrapURLs` can be set to `.remoteBootstrap` to load the css from other CDNs (though that option currently doesn't seem to functioning correctly as of c78ca96069732beef25feaab481d8cb2642e5aaa).

At least for now the canonical url is generated without any options, so the robustness should have increased with these changes.
https://github.com/twostraws/Ignite/blob/c78ca96069732beef25feaab481d8cb2642e5aaa/Sources/Ignite/Elements/Head.swift#L112

While the modified code still assumes the index.html in the root level of the site,
if it doesn't meet the conditions, it will be ignored and behave as it did previously.

Finally this PR only touches on the part that involves the site to run locally, and doesn't touch the core HTML generation, so the overall risk should be lower as well.

Let me know of your thoughts!

JPToroDev:
Thanks for the clarification—this looks to be in great shape. Nice work!

tockrock:
Thanks for reviewing this!
Appreciate you maitaining this project.

PR #819, opened 2025-05-02
Adding Site-Wide Search to Ignite

JPToroDev:
This PR offers a possible implementation of site-wide search leveraging [lunr.js](https://lunrjs.com/), which integrates with Ignite with a single JS file, much like Prism. During publishing, we use the `title`, `description`, and `url` properties of all pages—and also the `tags` and `date` properties of articles—to build a search index that lunr queries via its JS.

Pages that lack a description will be ignored and emit a warning during publishing.

The API looks like this:

```swift
SearchForm { result in
    result.title
        .font(.title4)
        .fontWeight(.bold)
        .foregroundStyle(.bootstrapPurple)
    result.description
        .foregroundStyle(.black)
        .lineSpacing(1.2)
} noResultsView: {
    Text("No results were found!")
} resultsPageHeader: {
    Text("Looking for something?")
}
```

And looks like this in action:

Image

You can see site-wide search in action on [this branch of IgniteSamples](https://github.com/twostraws/IgniteSamples/tree/site-search).

### DIFF
--- a/Sources/Ignite/Elements/Body.swift
+++ b/Sources/Ignite/Elements/Body.swift
@@ -72,6 +72,11 @@ public struct Body: MarkupElement {
 
         output += Script(file: "/js/ignite-core.js").markup()
 
+        if publishingContext.isSearchEnabled {
+            output += Script(file: "/js/lunr.js").markup()
+            output += Script(file: "/js/search.js").markup()
+        }
+
         if isBoundByContainer {
             attributes.append(classes: ["container"])
         }
--- a/Sources/Ignite/Elements/SearchForm.swift
+++ b/Sources/Ignite/Elements/SearchForm.swift
@@ -1,0 +1,229 @@
+//
+// SearchForm.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+/// An action that triggers the search functionality.
+struct SearchAction: Action {
+    func compile() -> String {
+        "performSearch(document.querySelector('[id^=\'search-input-\']').value)"
+    }
+}
+
+/// A form that performs site-wide search.
+public struct SearchForm: HTML, NavigationItem {
+    /// The appearance of the search-button label.
+    public enum SearchButtonStyle: Sendable, Equatable, CaseIterable {
+        case iconOnly, titleAndIcon, titleOnly
+    }
+
+    /// The content and behavior of this HTML.
+    public var body: some HTML { self }
+
+    /// The standard set of control attributes for HTML elements.
+    public var attributes = CoreAttributes()
+
+    /// Whether this HTML belongs to the framework.
+    public var isPrimitive: Bool { true }
+
+    /// The view displayed for each search result.
+    private var resultView: any HTML
+
+    /// The view displayed when there are no results.
+    private var noResultsView: any HTML
+
+    /// A view displayed at the top of the search results page.
+    private var resultsPageHeader: any HTML
+
+    /// This text provides a hint to users about what they can search for.
+    private var prompt: String = "Search"
+
+    /// How a `NavigationBar` displays this item at different breakpoints.
+    public var navigationBarVisibility: NavigationBarVisibility = .automatic
+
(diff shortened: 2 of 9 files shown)


twostraws:
Conceptually the idea of site-wide search is great, and would be a welcome addition. I have my usual concerns about adding a third-party dependency, particularly for filtering based on only three properties, but putting that to one side this feels like an odd API shape. If this were a regular Ignite page type (alongside `StaticPage` and `ArticlePage`) then users could design it however they wanted, and access a `searchResults` property via protocol extension that provides an array of each article's title, description, URL, image, tags, etc.

JPToroDev:
> particularly for filtering based on only three properties

The implementation uses only three properties for standard pages—and five properties for articles—to prevent performance issues. A search index is a JSON file hosted on the website. Say a website has 100 searchable pages: indexing three properties that naturally have the highest concentration of keywords creates a much smaller index file than indexing the entire body of each page—which would create an index file that is essentially 100 pages in size.

This "tradeoff" isn't unique to lunr.js, but to client-side search in general. That said, I think the solution—encouraging folks to write highly searchable titles and descriptions—is a good thing.

> this feels like an odd API shape. If this were a regular Ignite page type

I had considered making a `SearchPage` type initially, but that API design is at odds with how the HTML needs to be generated.

The view used for a search result needs to be defined by itself, because it creates a reusable `<template>` element. 

Something like this isn't possible:

```swift
struct MySearchPage: SearchPage {
    var body: some HTML {
        Text("Search Results")
        ForEach(searchResults) {
            result.title
        }
    }
}
```

Because we need to generate HTML that looks like this for the search result:

```HTML
<template>
    <p class="result-title"></p>
</template>
```

But we don't know where the reusable template starts and ends in the view tree of `MySearchPage`. 

To make a `SearchPage` type work, we'd need an API like this:


```swift
struct MySearchPage: SearchPage {

    func resultView(result: SearchResult) → some HTML {
         result.title
            .font(.title4)
            .fontWeight(.bold)
            .foregroundStyle(.bootstrapPurple)
        result.description
            .foregroundStyle(.black)
            .lineSpacing(1.2)
    }

    var body: some HTML {
        ForEach(searchResults) { // implicit property, but this property could also be used in the above method, which shouldn't have access to it
            result.title
        }
    }
}
```

Or this:

```swift
struct MySearchPage: SearchPage {

    func resultView(result: SearchResult) -> some HTML {
         result.title
            .font(.title4)
            .fontWeight(.bold)
            .foregroundStyle(.bootstrapPurple)
        result.description
            .foregroundStyle(.black)
            .lineSpacing(1.2)
    }

    func body(searchResults: [SearchResultView]) -> some HTML {
        ForEach(searchResults) {
            result.title
        }
    }

    func noResultsView() -> some HTML {
        Text("No results...")
    }
}
```

Those API shapes look like nothing else in Ignite.

What's more, the current API allows people to create different search pages for each instance of `SearchForm`. That wouldn't be possible if users were to add a single `SearchPage` conforming type to a `searchPage: any SearchPage` requirement of `Site`.

MrSkwiggs:
> Something like this isn't possible:
>
> ```swift
> struct MySearchPage: SearchPage {
>     var body: some HTML {
>         Text("Search Results")
>         ForEach(searchResults) {
>             result.title
>         }
>     }
> }
> ```
>
> Because we need to generate HTML that looks like this for the search result:
>
> ```HTML
> <template>
> <p class="result-title"></p>
> </template>
> ```
>
> But we don't know where the reusable template starts and ends in the view tree of `MySearchPage`.

Could we perhaps specialise `ForEach` when contained in a `SearchPage`-conforming type? Or override its declaration in `SearchPage`-conforming types? And wrap its content with `<template>` anchors.

This would be somewhat transparent to the user while giving us all the necessary control required for this sort of behaviour 🤔

JPToroDev:
If we wanted to use a `ForEach`, perhaps something like this would be our best bet?

```swift
struct MySearchPage: SearchPage {
    var body: some HTML {
        Text("Search Results")
        ForEach(searchResults) {
            SearchResult {
                result.title
            }
        }
    }
}
```

PR #820, opened 2025-05-03
Fix the standard remote CSS that's being called

tockrock:
The remoteIconCss is also being called below
https://github.com/twostraws/Ignite/blob/c78ca96069732beef25feaab481d8cb2642e5aaa/Sources/Ignite/Elements/Head.swift#L106-L107

The .remoteBootstrap option may not be handled correctly in other places, so this is not intended to be a a full fix, but more as a step in the right direction


tockrock:
Thanks for taking over!

PR #843, opened 2025-06-25
Adds more detail to doc comment for Image `init(_, description:)`

theMikeSwan:
This adds info on now to supply dark mode and 2x variants for images. For folks that do not spend much time building web pages it was not clear until poking around in the code how to set things up so that an image varies based on light versus dark mode. This aims to make it a bit easier for users to know what to do.

It also updates the first parameter name to match the signature.

### DIFF
--- a/Sources/Ignite/Elements/Image.swift
+++ b/Sources/Ignite/Elements/Image.swift
@@ -32,8 +32,10 @@ public struct Image: InlineElement, LazyLoadable {
     /// Creates a new `Image` instance from the specified path. For an image contained
     /// in your site's assets, this should be specified relative to the root of your
     /// site, e.g. /images/dog.jpg.
+    /// Append `~dark` to the end of filenames for a dark mode version of the image. (`cool-image.svg` and `cool-image~dark.svg`)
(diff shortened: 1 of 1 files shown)


MrSkwiggs:
I believe this functionality should actually be implemented and handled automatically by Ignite if we can detect light/dark mode or scaled assets. It will require some work though. 

If you want to give it a shot, you can do so from within this PR. Otherwise, we'll create a ticket for this to be tackled by the team.

theMikeSwan:
This functionality is already in place, `findVariants(for path: String) -> (light: [URL], dark: [URL])` takes care of sorting out the different variants and has a doc comment that talks about adding `~dark`, `@2x`, or even `@2x~dark`. That just isn't a method most folks are going to come across so I was thinking it would be good to move that info up a little higher.
My next stop is the Ignite examples to add a note there too.

I should note that the dark vs. light image is determined by the system setting, not which theme gets picked. It is possible to be in dark mode on your machine, select a light theme for the site, and, in my case, have the image effectively vanish.

twostraws:
Sorry for taking so long to get around to this! It looks good to me – thank you!

PR #844, opened 2025-06-25
Ignite v0.6.5: Introduce Variadic Generics for Expressive Typing

jptorodev:
This PR completes the expressive typing work begun in 0.6.0 by introducing variadic generics (SwiftUI's approach) to the framework.

While this primarily affects internal types, several surface API changes were unavoidable:

### Grid API Redesign
The biggest change is to the `Grid` API, whose previous implementation had become unmaintainable.

The new API is modeled after SwiftUI, uses CSS Grid instead of Bootstrap, and leverages deprecation messages for non-breaking changes.

Before:
```swift
Grid {
    Text("1")
        .width(1)
    Text("2")
        .width(1)
}
.columns(2)
```

After:
```swift
Grid(columns: 2) {
    Text("1")
        .gridCellColumns(1)
    Text("2")
        .gridCellColumns(1)
}
```

However, `Grid` without modifiers now behaves differently:
```swift
Grid {
    Text("1")
    Text("2")
}
```

Like SwiftUI, each `Text` now takes up its own full row.

For a complete overview of the new `Grid` API, visit: https://github.com/JPToroDev/IgniteSamples/blob/current/Sources/Pages/Concepts/GridExamples.swift

### `tagLinks()` Return Type Change
Since `Link` is now generic, `Article/tagLinks()` returns `some HTML` instead of `[Link]`:

Before:
```swift
let tagLinks = article.tagLinks()
if let tagLinks {
    Section {
        ForEach(tagLinks) { link in
            link
        }
    }
}
```

After:
```swift
Section {
    article.tagLinks()
}
```

To simplify manual `Link` construction, `Article/links` now returns `[TagMetadata]` instead of `[String]`. `TagMetadata` contains both the tag name and URL.

### `font()` Modifier for `Text`
To change the semantic HTML tag of `Text` elements (`<p>`, `<h1>`, etc.), ensure `.font()` is the first modifier applied. Otherwise, it will only affect font size without changing the underlying HTML tag. Introducing environment modifiers will remove this restriction in the future.

### Protocol and Type Naming Changes
1. Protocol names changed following the pattern: `NavigationItem` → `NavigationElement`. These shouldn't be directly conformed to, so impact should be minimal.

2. Nested types were flattened: `Card.Style` → `CardStyle`. This should affect few users.

## On Variadic Generics

Previously, Ignite erased almost everything to `AnyHTML`, losing important type information. This made it difficult to handle type-specific behaviors in complex view hierarchies.

Variadic generics preserve type information by allowing wrapper types to conditionally inherit protocol conformances from their content. For example, when a `Link` (which conforms to `LinkProvider`) is wrapped in modifiers, the resulting `ModifiedHTML` type also conforms to `LinkProvider`.

This enables better type safety and more sophisticated compile-time optimizations while maintaining SwiftUI-like ergonomics. The upgrade required several new internal types: `HTMLModifier`, `PackHTML` (equivalent to SwiftUI's `TupleView`), and new result builders.

### DIFF
--- a/Sources/Ignite/Elements/TableColumn.swift
+++ b/Sources/Ignite/Elements/TableColumn.swift
@@ -5,7 +5,7 @@
 // See LICENSE for license information.
 //
 
-@available(*, deprecated, message: "Column has been renamed to TableColumn")
+@available(*, deprecated, message: "Column has been renamed to TableColumn and will be removed in a future release.")
 public typealias Column = TableColumn
 
 /// A column inside a table row.
--- a/Sources/Ignite/Elements/TableRow.swift
+++ b/Sources/Ignite/Elements/TableRow.swift
@@ -5,7 +5,7 @@
 // See LICENSE for license information.
 //
 
-@available(*, deprecated, message: "Row has been renamed to TableRow")
+@available(*, deprecated, message: "Row has been renamed to TableRow and will be removed in a future release.")
 public typealias Row = TableRow
 
 /// One row inside a `Table`.


.github/workflows/swift.yml
```diff
@@ -1,21 +1,20 @@
...
   build:
-
     runs-on: macos-15
-
     steps:
     - uses: actions/checkout@v4
+    - name: Setup Xcode
+      uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: 'latest-stable'
```
MrSkwiggs, line 17: This needed?

Sources/Ignite/Elements/Abbreviation.swift
```diff
@@ -6,24 +6,21 @@
 //
 
 /// Renders an abbreviation.
-public struct Abbreviation: InlineElement {
+public struct Abbreviation<Content: InlineElement>: InlineElement {
     /// The content and behavior of this HTML.
-    public var body: some InlineElement { self }
+    public var body: Never { fatalError() }
```
MrSkwiggs, line 11: Hmm I wonder if there isn't a better way to handle this? Couldn't we just use `var body: Content { content }` instead? and do we even need `var content: Content` even? 

I may have missed a critical discussion about this as I see it was already in place, so feel free to let me know if that is the case

Sources/Ignite/Framework/ElementTypes/InlineElement.swift
```diff
@@ -8,29 +8,69 @@
 /// An element that exists inside a block element, such as an emphasized
 /// piece of text.
 @MainActor
-public protocol InlineElement: BodyElement, CustomStringConvertible {
+public protocol InlineElement: CustomStringConvertible, Sendable {
```
MrSkwiggs, line 11: Does this work?
```suggestion
public protocol InlineElement: CustomStringConvertible, Sendable, HTML where Body: InlineElement {
```

JPToroDev:
> > @ralfboernemeier If you'd be willing: could you give the new [type-preservation](https://github.com/twostraws/Ignite/tree/type-preservation) branch a try and let me know if you encounter any issues using the update guide above?
> 
> I would like to try it, but I currently have very little free time in my spare time. I’ll try it as soon as I have time, but unfortunately I can’t promise anything - sorry.

No pressure. Keeping you in the loop to make sure we address your concerns!

Sources/Ignite/Elements/Abbreviation.swift
JPToroDev: This approach mirrors SwiftUI. Primitive types almost never use their `body`, so making it return `Never` easily allows us to determine which elements are custom user elements, and which elements belong to the framework, encapsulated in the computed `isPrimitive` property.

Sources/Ignite/Framework/ElementTypes/InlineElement.swift
JPToroDev: `InlineElement` cannot inherit from `HTML` because the former uses `@InlineElementBuilder` and the latter uses `@HTMLBuilder` for their `body` vars.

ralfboernemeier:
@JPToroDev After I was able to solve the Table() problem, it was with the help of the explanations here in the ticket, in my case in particular the changes to Grid, very easy to change my website in the branch and compile it successfully. I find the idea with the examples - before and after - very helpful and intuitive!

ralfboernemeier:
@JPToroDev May I ask a question about what I consider to be a breaking change? Ich habe folgenden Code:
```
Grid(columns: 4) {
            Text {
                Link(Image("/images/1.svg", description: "Claude").resizable().frame(width: 80, height: 80), target: "https://claude.ai")
                    .target(.blank)
                    .role(.secondary)
            }
            .gridCellColumns(1)
            .horizontalAlignment(.center)
            
            Text {
                Link(Image("/images/2.svg", description: "Speed").resizable().frame(width: 80, height: 80), target: Speedtest())
                    .target(.parent)
                    .role(.secondary)
            }
            .gridCellColumns(1)
            .horizontalAlignment(.center)
            
            Text {
                Link(Image("/images/3.svg", description: "Fritz!").resizable().frame(width: 80, height: 80), target: "http://target2.html")
                    .target(.blank)
                    .role(.secondary)
            }
            .gridCellColumns(1)
            .horizontalAlignment(.center)
            
            Text {
                Link(Image("/images/4.svg", description: "Cloud").resizable().frame(width: 80, height: 80), target: "https://target4.html")
                    .target(.blank)
                    .role(.secondary)
            }
            .gridCellColumns(1)
            .horizontalAlignment(.center)
            
        }
        .padding(.top, 20)
```
With 0.6.0, the 4 images are displayed in a row on my mobile device, i.e. next to each other. With the branch, the images on my mobile device are displayed one below the other, i.e. 4 rows below each other. It does not matter whether I reduce the size of the images to ‘20’, for example. On a PC, the icons in both versions are displayed next to each other in one row, as desired. 
Is there an explanation for this, or even better, a solution?

JPToroDev:
Adding `gridItemSizing(.adaptive(minimum: 0))` will ensure the grid items remain in the same row on mobile.

JPToroDev:
Thank you for giving this branch a whirl and for offering all the great feedback!

Sources/Ignite/Elements/TableRow.swift
```diff
@@ -5,8 +5,11 @@
 // See LICENSE for license information.
 //
 
+@available(*, deprecated, message: "Row has been renamed to TableRow")
+public typealias Row = TableRow
+
```
MrSkwiggs, line 10: Ditto here

Sources/Ignite/Elements/TableColumn.swift
```diff
@@ -5,8 +5,11 @@
 // See LICENSE for license information.
 //
 
+@available(*, deprecated, message: "Column has been renamed to TableColumn")
+public typealias Column = TableColumn
+
```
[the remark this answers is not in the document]
JPToroDev: > Since we're not at 1.0.0 yet, I'd argue we should not add deprecation here and just perform the rename.

I agree with you, but Paul has advocated for these deprecation warnings based on feedback he has received. I've improved the warnings to mention removal in a future release.

PR #849, opened 2025-09-09
Add type attribute to Script to be used in HTML head.

mwermeester:
Currently, there is no way to specify a `type` attribute for the `Script` header element.

I want to add _JSON-LD_ to my site and the recommended way is to add it in the `Script` element by specifying the the `type` attribute `application/ld+json`.

This pull request might not be the best way to implement it, but at least it provides a workaround.


twostraws:
Thank you!

PR #852, opened 2025-11-05
Fix: `Style` class name generation produces invalid identifiers, breaking style application

### LINKED ISSUE #851: Incorrect class name generation for `Style` causes styles not to apply
Due to the changes introduced #816 in commit 575bcc27f60af06a9e3ff04a4b2b2ef66274f3cc￼, the way class names related to `Style` are generated has been modified.

The previous implementation can be seen here:

https://github.com/twostraws/Ignite/blob/888e95c65fd1ae2daf47cb9712885d73fffbe3fe/Sources/Ignite/Publishing/CSS/StyleManager.swift#L46-L61

And the new implementation after commit 575bcc27f60af06a9e3ff04a4b2b2ef66274f3cc is as follows:

https://github.com/twostraws/Ignite/blob/575bcc27f60af06a9e3ff04a4b2b2ef66274f3cc/Sources/Ignite/Styles/Style.swift#L24-L43

In both versions, the base string for generating the class name is created using `String(describing: type(of: style))`.
However, in the newer implementation, it seems that `style` should actually refer to `self`.
Currently, regardless of the concrete `Style` type, because `Style.style(content:environment:)` is used, the result becomes:

```swift
"@MainActor (StyledHTML, EnvironmentConditions) -> StyledHTML"
```

This string includes characters that cannot be used in a class name (`@main-actor-(styled-html,-environment-conditions)-->-styled-htmlstyle`), which causes the style not to be applied properly.

treastrain:
Fix #851

### DIFF
--- a/Sources/Ignite/Styles/Style.swift
+++ b/Sources/Ignite/Styles/Style.swift
@@ -35,7 +35,7 @@ extension Style {
     /// The name of the CSS class this `Style` generates,
     /// derived from the type name minus the "Style" suffix, if present.
     var className: String {
-        let typeName = String(describing: type(of: style))
+        let typeName = String(describing: type(of: self))
         let baseName = typeName.hasSuffix("Style") ? typeName : typeName + "Style"
         let className = baseName.kebabCased()
         return className


treastrain:
@JPToroDev @twostraws 

I have fixed a bug introduced in #816 (commit 575bcc27f60af06a9e3ff04a4b2b2ef66274f3cc). Could you please review this change?

PR #853, opened 2025-11-12
Feature request: Add `hoverEffect(_:)` support for `InlineElement`

treastrain:
Due to commit eb9136db2ddfa6488caef1ee1bea2d5e2575b1a in #142, `hoverEffect(_:)` can no longer be called on the current `InlineElement` due to a missing protocol conformance.

This pull request restores support for `hoverEffect(_:)` on `InlineElement`.

### DIFF
--- a/Sources/Ignite/Modifiers/HoverEffect.swift
+++ b/Sources/Ignite/Modifiers/HoverEffect.swift
@@ -10,7 +10,7 @@ public extension HTML {
     /// - Parameter effect: A closure that returns the effect to be applied.
     ///   The argument acts as a placeholder representing this page element.
     /// - Returns: A modified copy of the element with hover effect applied
-    func hoverEffect(_ effect: @escaping (EmptHTMLyHoverEffect) -> some HTML) -> some HTML {
+    func hoverEffect(_ effect: @escaping (EmptyHTMLHoverEffect) -> some HTML) -> some HTML {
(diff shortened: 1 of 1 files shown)


treastrain:
@twostraws Thank you very much for your review and your comment!

Regarding the rename from `EmptyHoverEffect` to `EmptyHTMLHoverEffect`, I made this change because the original `EmptyHoverEffect` stopped conforming to the `InlineElement` protocol after #142 and now conforms only to `HTML`. This was one of the main reasons I created this pull request.

I initially considered extending the existing `EmptyHoverEffect` so that it would conform to both `HTML` and `InlineElement`. However, both protocols define an associated type `Body` and a `body` property with the same name but incompatible requirements: the `HTML` version uses the `@HTMLBuilder` result builder and must return an `AnyHTML` value, whereas the `InlineElement` version uses the `@InlineElementBuilder` result builder and must return `some InlineElement`.

Because of these conflicting constraints, a single type cannot satisfy both protocols simultaneously. I also considered whether the `@_implements` attribute could help bridge this gap, but it turned out not to be applicable in this particular case.

For these reasons, I introduced two separate structs—`EmptyHTMLHoverEffect` and `EmptyInlineElementHoverEffect`—so that each can conform appropriately to its respective protocol. I hope this clarifies the reasoning behind the rename, and I am happy to revise the naming if you would prefer a different approach.

PR #859, opened 2026-01-06
fix #858 RunCommand: find server script by CLI; report path in error message

### LINKED ISSUE #858: Run command ignores PREFIX_DIR installs
**Short description**
To install in a different directory, the README suggests

   `make install PREFIX_DIR=/my/install/dir`

But after such an install, `ignite run` only seeks `ignite-server.py` in `/usr/bin/local`, and emits an error.

Also, the error message is reversed (saying the fix is `make install && make clean`).

**Steps to Reproduce**
1.   `make install PREFIX_DIR=/my/install/dir`
2. ignite run

RESULT: error message: `... make install && make clean `

**Expected behavior**
- Find (and launch) python script next to tool
- When not found, issue error with path and correct make command

**Build information**
 - OS: macOS 26.1
 - Swift version:  Swift 6.2
 - Ignite version: latest main branch 
 - Tested browsers: n/a

wti:
Fixes #858.

- Uses ProcessInfo argument 0 (always available) to derive the install dir
- Includes the missing path in the error message
- Suggests make clean before install instead of vice-versa

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -18,5 +18,7 @@ jobs:
     - uses: actions/checkout@v4
     - name: Build
       run: swift build
-    - name: Run tests
-      run: swift test -q
+    - name: Run tests (except SiteTests)
+      run: swift test -q --skip SiteTests
+    - name: Run tests (SiteTests only)
+      run: swift test -q --filter SiteTests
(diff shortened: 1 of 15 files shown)


wti:
wrt failing test "Site published when Markdown content contains invalid lastModified date", there may be a filesystem race.

The test writes a file and then tries to detect it.  This also failed on my machine initially on a clean tree.  I tried changing the file-write to be atomic, which worked, but the test also worked after changing it back (I haven't investigated the test cleanup, but it seems to be skipped on failure).  I can think of no way that the RunCommand changes trigger this failure.

I'm not set up to replicate CI, but next steps on the test could be to do atomic writes and always clean up (using something like `defer{ cleanup... }`).

PR #863, opened 2026-02-03
In CI, run SiteTests in isolation as recommended

wti:
Configure swift.yml CI tests to run `SiteTests` separately from other tests. 

PR's #861 and #859 are blocked by a failing test.

[Tests/IgniteTesting/Publishing/Site.swift](https://github.com/twostraws/Ignite/blob/58b5e1c7f22d188ed6d5d8df5d9d76c1e805db03/Tests/IgniteTesting/Publishing/Site.swift#L15) says the tests should be run in isolation because they set context shared with other tests.  (The tests are serialized in the main thread).

I was able to reproduce the problem locally when running tests together, but unable to reproduce when running separately, like this CI configuration.


wti:
The key seems to be generating the `Content` dir.  Only two tests created files assuming that the Content dir existed (a third created it).  If I put the other test first, it also failed unless the Content dir was created.

These tests could use more work (the `Tests` subdirectories are not guaranteed writable), but I hope they are more reliable now.

Issue #864, opened 2026-02-22
Support full, remote URL for Images.


vdhamer:
### Problem

The examples website states
> Images can be either pictures from your /Assets/images folder, or one of the built-in icons.

I am in need of an option where the generated HTML images are simply loaded from a full, remote URL.

### Why?

My static website is generated by software from a "database" (actually a bunch of version controlled JSON files). So one data source to my app is a set of full imageURLs. With Ignite, I currently need to move these to the /Assets/images folder. But I can get name clashes if two URLs have very different paths, but have the same filename: the HTML page will display 2 identical images because the first one got overwritten. Subdirectories within images/ could solve this, but gets messy because any component of the path may be essential.

### Solution

Provide an extra variant of Image() that accepts a full URL. And generate the HTML to fetch that image simply using that full URL.

Discussion: if the link breaks because the host disappears of restructures, we won't see the desired image.
But for my application this is OK: the URL is not used to "borrow" someone else's image to embellish my site.
The other party is aware that their image is used, and actually has to select which image they want used. So they provide a full URL, and if it changes, they need to adapt the URL.

### Short term workaround (for my own code)

I plan to append a suffix as soon as there is a filename clash: "myImage.jpg", "myImage_2.jpg, etc.
This approach is used more often (e.g. Import feature of Adobe Lightroom may find a name clash when adding images for a folder). It involves tracking the suffix (potentially empty string) in a dictionary while the site is generated. A bit more complicating than "HTML allows remove URLs, so you can provide a remote URL".

MrSkwiggs:
Hi @vdhamer, thanks for bringing this up. While it may seem that loading remote images is not supported, it actually is! 

You can try this for yourself with the following very simple example:

```swift
struct RemoteImagePage: StaticPage {
    var title = "Remote Image"
    
    var body: some HTML {
        VStack(alignment: .leading) {
            Text("Remote Image")
                .font(.title1)
            Text("This page shows how to load an image from a remote URL.")
            Spacer()
            Image("https://picsum.photos/200")
        }
    }
}
```

Which produces the following:
<img width="531" height="309" alt="Image" src="https://github.com/user-attachments/assets/0dc41cf3-71e4-42cf-8f63-7571b60a890c" />

Now, granted the logs may make it seem like Ignite is not able to locate the image from the given path, but the generated HTML is valid and works as it should.

For now, I'd recommend simply using the Image element with remote URLs as you initially needed to, while we improve this internally 👍

[...]

PR #865, opened 2026-02-25
Skip local variant detection for remote URL images

william-laverty:
## Summary

When an `Image` is initialized with a remote URL (`http`/`https`), skip the local filesystem variant detection (`findVariants`) that searches for `@2x` and `~dark` alternatives in the assets directory.

## Problem

As noted in #864, remote URLs already work with `Image`, but Ignite attempts to find local file variants for them — scanning the assets directory for density and appearance alternatives. This is unnecessary for remote images and produces warnings when the assets directory doesn't contain matching files.

## Fix

- Add an `isRemote` flag to `render(path:description:)` that skips `findVariants()` entirely for `http`/`https` URLs.
- Detect remote URLs in `markup()` by checking the URL scheme.

This is a minimal, targeted change — remote images emit a clean `` tag without triggering filesystem operations.

### DIFF
--- a/Sources/Ignite/Elements/Image.swift
+++ b/Sources/Ignite/Elements/Image.swift
@@ -97,12 +97,18 @@ public struct Image: InlineElement, LazyLoadable {
     /// - Parameters:
     ///   - path: The user image to render.
     ///   - description: The accessibility label to use.
+    ///   - isRemote: Whether this is a remote URL (skips local variant detection).
     /// - Returns: The HTML for this element.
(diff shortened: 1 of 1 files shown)


MrSkwiggs (approved):

MrSkwiggs:
Thanks for your contribution @William-Laverty! 
I've elected to disable automatic resolution of #864 for now, as that issue was happening for an older release of Ignite. 

Your changes look good to me however, so I'm happy to merge this. 
We'll close #864 manually once OP gets a chance to report back.

PR #866, opened 2026-02-26
feat: Support remote image URLs

WZBbiao:
This PR adds support for remote image URLs (http/https) in the `Image` element.\n\nCurrently, `Image` assumes all paths are local assets and tries to resolve them relative to the site root or find variants in the assets directory. This change detects if the path starts with `http://` or `https://` and renders a standard `` tag without attempting to resolve local paths or variants.\n\nCloses #864


MrSkwiggs:
Thanks for your contribution @WZBbiao. As an earlier PR addressing the exact same issue was submitted before yours, I am closing this. If the other PR ends up not being merged, please feel free to reopen.

PR #869, opened 2026-03-11
Fix asset copying to merge with existing directories

jpurnell:
## Summary

This PR fixes a crash that occurs when users have folders in `Assets/` with the same names as Ignite-generated folders (e.g., `css/`, `fonts/`, or any page name).

**The problem:**
- `copyAssets()` and `copyFonts()` use `FileManager.copyItem()` which throws if the destination already exists
- Ignite generates `css/` and `fonts/` folders during content generation (for bootstrap, ignite-core, etc.)
- When users have `Assets/css/custom.css`, the copy fails because `docs/css/` already exists
- `clearBuildFolder()` uses `try?` and can silently fail with file locks (especially on Dropbox)

**The fix:**
- Adds a recursive `mergeContents(from:to:)` helper that:
  - For files: removes existing and copies (allowing overwrites)
  - For directories: creates if needed, then recursively merges contents
- Updates `copyAssets()`, `copyFonts()`, and `copy(resource:)` to use this helper

**Result:**
Users can now have `Assets/css/custom.css` and it will appear alongside Ignite's `bootstrap.min.css` in the final `docs/css/` folder.

## Test plan

- [x] Verified fix compiles
- [x] Tested with a project that has `Assets/css/`, `Assets/fonts/`, and `Assets/constitution/` folders
- [x] Confirmed merged output contains both Ignite files and user files
- [x] Tested multiple builds without cleaning (no more intermittent failures)


twostraws:
Sorry for taking so long to read through this – thank you!

PR #870, opened 2026-03-15
Set data-light-theme and data-dark-theme on html element

jpurnell:
## Summary

- **Fixes custom theme dark mode switching** by setting `data-light-theme` and `data-dark-theme` attributes on the `<html>` element in `PlainDocument.markup()`
- The existing theme-switching JavaScript reads these attributes to determine which CSS selectors to activate, but they were never set — causing custom themes to fall back to `"light"`/`"dark"` which don't match custom theme CSS selectors like `[data-bs-theme="my-theme-light"]`
- Adds two tests verifying the attributes are present with correct cssIDs

## Test plan

- [x] New tests in `HTMLDocumentTests` verify `data-light-theme` and `data-dark-theme` attributes are set on `<html>`
- [x] Full test suite passes (1075 tests, 0 failures)
- [ ] Verify with a site using custom themes that dark mode switching works without a JS workaround

### DIFF
--- a/Sources/Ignite/Elements/PlainDocument.swift
+++ b/Sources/Ignite/Elements/PlainDocument.swift
@@ -30,6 +30,14 @@ public struct PlainDocument: Document {
         var attributes = attributes
         attributes.append(customAttributes: .init(name: "lang", value: language.rawValue))
 
+        let site = PublishingContext.shared.site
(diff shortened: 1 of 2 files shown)


twostraws:
Thank you!

PR #871, opened 2026-03-16
Fix XML escaping in RSS feed generator

jpurnell:
## Summary
- Article titles and site metadata containing special XML characters (`&`, `<`, `>`, `"`, `'`) produced invalid RSS that feed readers rejected
- Added `xmlEscape()` helper to `FeedGenerator` to properly encode non-CDATA text fields (item titles, site name, site description)
- Fields already wrapped in `<![CDATA[...]]>` (description, author, tags, content) are unaffected

## Test plan
- Added test verifying `&` in article titles is escaped to `&amp;` in feed output
- All existing `FeedGenerator` tests continue to pass
- Validated output with `xmllint --noout` on a real feed containing titles with `&` characters

### DIFF
--- a/Sources/Ignite/Publishing/FeedGenerator.swift
+++ b/Sources/Ignite/Publishing/FeedGenerator.swift
@@ -17,6 +17,15 @@ struct FeedGenerator {
         self.content = content
     }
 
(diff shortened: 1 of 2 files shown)


twostraws:
Thank you!

PR #875, opened 2026-04-04
Fix Markdown block rendering in Text

### LINKED ISSUE #874: Text(markdown:) drops headings and can generate invalid HTML for block Markdown
`Text(markdown:)` currently treats rendered Markdown as either a single paragraph or multiple adjacent paragraphs and then does HTML string surgery on the result.

That breaks for block Markdown that mixes headings, paragraphs, and lists:
- the first heading is always parsed with `removeTitleFromBody: true`, so a heading can disappear from the rendered body
- block content can be rewrapped into invalid paragraph structure
- after HTML prettification this can show up as empty `<p></p>` elements that adversely affect layout

Minimal example:

```swift
Text(markdown: """
## Heading 1
Text 1

Text 2

## Heading 2
Text 3
""")

Expected:

- both headings are preserved
- paragraphs remain valid
- block Markdown renders as valid block HTML

Actual:

- headings can disappear
- invalid paragraph HTML can be produced
- prettification can normalize the broken structure into empty paragraphs

esconner:
Fixes #874

This changes `Text(markdown:)` and `Text(markup:parser:)` to preserve block Markdown structure instead of doing fragile HTML string
splitting.

What changed:
- stop stripping the first heading from `Text` Markdown bodies
- detect the true single-paragraph case from parsed HTML
- render block Markdown inside a container so headings/lists/paragraphs remain valid
- escape plain text and inline/code block contents in `MarkdownToHTML` so the renderer always produces valid HTML

Tests:
- preserve multiple headings in `Text(markdown:)`
- avoid invalid paragraph output for mixed block Markdown
- preserve block structure for headings + lists
- cover the same behavior through `Text(markup:parser:)`
- verify Markdown text/code escaping in `MarkdownToHTML`

Validation:
- validated in a separate Ignite-based project against real integration content and against IgniteSamples

### DIFF
--- a/Sources/Ignite/Elements/Text.swift
+++ b/Sources/Ignite/Elements/Text.swift
@@ -5,6 +5,8 @@
 // See LICENSE for license information.
 //
 
+import SwiftSoup
+
 /// A structured piece of text, such as a paragraph of heading. If you are just
 /// placing content inside a list, table, table header, and so on, you can usually
 /// just use a simple string. Using `Text` is required if you want a specific paragraph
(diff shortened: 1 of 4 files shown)


esconner:
Hi,

- I can apply the change you suggested. Do you generally prefer if and switch expressions when possible in Ignite?
- While I’m at it: do you prefer implicit returns too? If so I could also change my escapedHTML function in line 40.
- I’ll also run swiftlint to clear up some whitespace issues I caused (sorry).
- did you retract this comment from GitHub? I can’t see it there.

Cheers
Eric

PR #877, opened 2026-04-14
Add Atom and JSON Feed generators

jpurnell:
## Summary
- Add Atom (RFC 4287) and JSON Feed (v1.1) generators alongside the existing RSS 2.0 generator, giving site authors three industry-standard syndication options
- Introduce `FeedFormat` enum and extend `FeedConfiguration` with `formats` and per-format `paths`, fully backward compatible (defaults to `[.rss]` with existing behavior)
- Add `<link rel="alternate">` auto-discovery tags to `<head>` so feed readers can find feeds automatically
- Extract shared `String.xmlEscaped` and `Date.asRFC3339()` utilities used by both XML-based generators
- JSON Feed uses `Encodable` + `JSONEncoder` for safe serialization with no escaping bugs by construction
- 61 new and updated tests across 8 suites

## Test plan
- [x] All 61 feed-related tests pass (AtomFeedGenerator, JSONFeedGenerator, FeedConfiguration, FeedGenerator, FeedLink, Head, StringXMLEscaping, DateRFC3339)
- [x] Existing RSS generator tests unchanged and passing
- [x] Zero compiler warnings on `Ignite` target
- [x] Backward compatibility: existing sites with default `feedConfiguration` generate only RSS at `/feed.rss` with no code changes
- [x] Validate generated Atom output with [W3C Feed Validation Service](https://validator.w3.org/feed/)
- [x] Validate generated JSON Feed output against [jsonfeed.org spec](https://www.jsonfeed.org/version/1.1/) — all required fields present; added `content_text` fallback in descriptionOnly mode for strict compliance


MrSkwiggs:
Amazing work @jpurnell, thanks for submitting!

I see your checklist is still missing the final step, still something you want to handle before we proceed?

jpurnell:
Thanks for the catch! That final checkbox is now done.

Validated against the [JSON Feed v1.1 spec](https://www.jsonfeed.org/version/1.1/) — all required fields (`version`, `title`, `items`, item `id`) are present. I also found that the spec requires either `content_html` or `content_text` per item, and in `descriptionOnly` mode we only had `summary` (which the spec treats as distinct from content). Fixed in 98c7843 — `descriptionOnly` mode now sets `content_text` as a fallback so every item is strictly spec-compliant.

twostraws:
Hello! This is a great set of changes – thank you!

A couple of questions:

1. How is the new date extension different from the existing `Date.asISO8601` extension we have?
2. How is the XML escaping extension different from the existing `FeedGenerator.xmlEscape()` method?

Thank you!

jpurnell:
The date extension mostly just gives an offset for Time Zone…on my site, I like to have the time of the post, but it's UTC with the existing extension. I wasn't sure if changing the signature would break other people's work, and technically, Atom is looking for RFC3339, so thought it better to make that explicit.

XMLEscape is more flexible as a shared extension rather than private to the FeedGenerator in the RSS fix, so it can be used for the Atom Generator now too.

twostraws:
> The date extension mostly just gives an offset for Time Zone…on my site, I like to have the time of the post, but it's UTC with the existing extension. I wasn't sure if changing the signature would break other people's work, and technically, Atom is looking for RFC3339, so thought it better to make that explicit.

The current extension uses ISO 8601, but the specific defaults Apple chose are RFC3339. If you want to modify it to accept a time zone we can do that, but I think it should be the same extension rather than a second one. Would that be okay?


> XMLEscape is more flexible as a shared extension rather than private to the FeedGenerator in the RSS fix, so it can be used for the Atom Generator now too.

In that case, we should perhaps delete the version from `FeedGenerator`, and make it match the format in String-EscapedForJavaScript.swift, which would mean `escapedForXML()` – mostly because it keeps the namespace clear, given that we're extending all strings for everyone.

twostraws:
@jpurnell I'm just checking in: what should we do with this PR?

jpurnell:
Sorry…lot of life events. Made some changes:

  Date.asISO8601 — Folded the timezone capability into the existing extension as asISO8601(timeZone:). Kept var asISO8601 as a UTC convenience that delegates to the method, so existing callers (including Time.swift)
   keep working unchanged. Deleted the separate Date-RFC3339.swift.

  String.escapedForXML() — Renamed xmlEscaped to escapedForXML() to match the escapedForJavascript() namespace convention, made it a public extension, and consolidated all the RSS/Atom call sites onto it.

twostraws:
Thank you!

PR #878, opened 2026-04-24
Fix light mode secondary/tertiary colors rendering as white (#854)

### LINKED ISSUE #854: Light mode secondary color is white
**Short description**
Applying `.foregroundStyle(.secondary)` to a `Text` object makes it invisible in light mode.

**Steps to Reproduce**
1. Run `ignite new test` to create a new project.
2. Add the modifier `.foregroundStyle(.secondary)` to the default `Text("Hello world!")`.
3. Run `ignite build` and `ignite run`.

**Expected behavior**
In dark mode, the page renders as expected, with “Hello world!” in grey instead of white:
Image

In light mode, however, the text is rendered as white. In this screenshot I have highlighted the text in Safari to show it is still there but in white:

Image

**Build information**
 - OS: Ubuntu 24.04.3 LTS
 - Editor: VSCode 1.106.2 on macOS Tahoe 26.1
 - Swift version: Swift 6.2 86_64-unknown-linux-gnu
 - Ignite version: 0.6.0
 - Tested browsers: Safari 26.1 (21622.2.11.11.9), Chrome 142.0.7444.176

**Additional context**
Looking at the CSS, it seems the color is somehow set to 50 000+ which seems to be the source of the issue:
Image

Thanks for your help with this project, it has helped me create good-looking static websites with no web dev experience.

RomanticD:
Fixes #854.

## Summary

`.foregroundStyle(.secondary)` (and `.tertiary`) rendered as pure white in light mode, making text invisible on the default white background. The root cause is an initializer-overload mismatch in `Theme-DefaultImplementation.swift`:

```swift
// Before
var secondary: Color {
    colorScheme == .dark ?
    Color(red: 222, green: 226, blue: 230, opacity: 0.75) :
    Color(red: 33,  green: 37,  blue: 41,  opacity: 0.75)
}
```

`0.75` is a `Double`, not a `Percentage`, so Swift picks the **`Double` RGB initializer** that treats each channel as a 0–1 float and multiplies by 255:

```swift
public init(red: Double, green: Double, blue: Double, opacity: Double = 1) {
    let intRed   = Int(red   * 255)  // 33 * 255 = 8415
    let intGreen = Int(green * 255)  // 37 * 255 = 9435
    let intBlue  = Int(blue  * 255)  // 41 * 255 = 10455
    ...
}
```

The emitted CSS for `--bs-secondary` becomes `rgb(8415 9435 10455 / 75%)`. Browsers clamp each channel to `255`, so the color resolves to white (reported in the issue as "50 000+" channel values — same bug, seen through the devtools rounding).

Dark mode happened to look OK because white text on a dark background is still readable (it just wasn't the intended 75%-opacity light grey).

## Fix

Switch the four affected call sites from `Double` opacities to `Percentage`:

```swift
// After
Color(red: 222, green: 226, blue: 230, opacity: 75%)
Color(red: 33,  green: 37,  blue: 41,  opacity: 75%)
Color(red: 222, green: 226, blue: 230, opacity: 50%)
Color(red: 33,  green: 37,  blue: 41,  opacity: 50%)
```

`75%` / `50%` produce a `Percentage`, which is the only type that matches the integer-RGB initializer's `opacity:` parameter. The channels stay `33/37/41` and `222/226/230`, matching the inline-comment intent (Bootstrap's standard light/dark text colors).

This touches only `Theme-DefaultImplementation.swift` — the default `secondary` and `tertiary` Bootstrap color variables. Custom themes that override `secondary`/`tertiary` are unaffected. No behavior change in dark mode beyond finally hitting the intended 75%/50% opacity values instead of white.

## Test plan

- [x] Added regression tests in `Tests/IgniteTesting/Themes/ThemeDefaultColors.swift` that assert `DefaultLightTheme().secondary` / `DefaultDarkTheme().secondary` / `.tertiary` resolve to the expected RGB + opacity values. The tests fail on `main` (e.g. `color.red → 8415`) and pass with this change.
- [x] `swift test` — all 1124 tests pass.
- [x] `swiftlint lint` — no issues on touched files.
- [x] `swift build` — clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/Ignite/Themes/Theme-DefaultImplementation.swift
+++ b/Sources/Ignite/Themes/Theme-DefaultImplementation.swift
@@ -54,14 +54,14 @@ public extension Theme {
 
     var secondary: Color {
         colorScheme == .dark ?
-        Color(red: 222, green: 226, blue: 230, opacity: 0.75) :
-        Color(red: 33, green: 37, blue: 41, opacity: 0.75)
+        Color(red: 222, green: 226, blue: 230, opacity: 75%) :
+        Color(red: 33, green: 37, blue: 41, opacity: 75%)
     }
 
     var tertiary: Color {
         colorScheme == .dark ?
-        Color(red: 222, green: 226, blue: 230, opacity: 0.5) :
-        Color(red: 33, green: 37, blue: 41, opacity: 0.5)
+        Color(red: 222, green: 226, blue: 230, opacity: 50%) :
+        Color(red: 33, green: 37, blue: 41, opacity: 50%)
     }
 
     var background: Color {
--- a/Tests/IgniteTesting/Themes/ThemeDefaultColors.swift
+++ b/Tests/IgniteTesting/Themes/ThemeDefaultColors.swift
@@ -1,0 +1,57 @@
+//
+// ThemeDefaultColors.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+import Foundation
+import Testing
+
+@testable import Ignite
+
+/// Tests for the default `Theme` colors that rely on `colorScheme` to pick
+/// a light- or dark-mode value. Regression coverage for #854, where the
+/// default `secondary` and `tertiary` colors were rendered as white because
+/// the call sites passed `opacity: 0.75`/`0.5` (a `Double`) instead of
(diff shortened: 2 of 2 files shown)


twostraws:
Thank you!

PR #882, opened 2026-05-19
Add StructuredData element for JSON-LD structured data

jpurnell:
## Summary

Adds a new `StructuredData` head element that renders JSON-LD `<script type="application/ld+json">` tags for [Schema.org](https://schema.org) structured data.

- **Generic initializer** — any schema type with a properties dictionary
- **Raw JSON-LD** — pass a pre-formatted JSON string for full control
- **Convenience factories** — `Organization`, `WebSite`, `Event`, `Article`, and `BreadcrumbList` with typed parameters
- `Article` and `BreadcrumbList` auto-populate from the publishing context (article metadata, page URL, site URL)
- Conforms to `HeadElement` and `Sendable`; 51 tests; DocC catalog entry

## Usage

`StructuredData` is a `HeadElement`, so you use it inside a `Head { }` block in your Layout — the same place you'd put a `MetaTag`, `MetaLink`, or `Script`:

```swift
struct MyLayout: Layout {
    var body: some Document {
        Head {
            // Generic schema — any Schema.org type
            StructuredData("LocalBusiness", properties: [
                "name": "Joe's Pizza",
                "telephone": "555-0123"
            ])

            // Convenience methods
            StructuredData.organization(name: "Acme Corp", url: "https://acme.com")
            StructuredData.webSite(name: "Acme Corp", url: "https://acme.com")
            StructuredData.breadcrumbs()
            StructuredData.article(publisher: "Acme Corp", publisherURL: "https://acme.com")
            StructuredData.event(
                name: "Annual Conference",
                startDate: "2026-06-01",
                endDate: "2026-06-03",
                locationName: "Convention Center",
                locality: "Springfield",
                region: "IL",
                postalCode: "62701"
            )

            // Raw JSON-LD for schemas without a convenience method
            StructuredData(json: customJSONString)
        }

        Body {
            content
        }
    }
}
```

No site-level configuration is needed. Each `StructuredData` element renders an independent `<script type="application/ld+json">` tag in the page's `<head>`, and you can include as many as you need per page. Context-aware schemas like `Article` and `BreadcrumbList` automatically emit nothing when they don't apply (e.g., breadcrumbs on the homepage, article schema on non-article pages).

## Test plan

- [x] 51 tests covering all schema types, edge cases, unicode, invalid input, stress tests, and documentation examples
- [x] Builds clean on Swift 6.3 with strict concurrency (`Sendable` conformance, no `@MainActor`)
- [x] Tests use the `.publishingContext()` trait (current upstream pattern)
- [x] JSON-LD output validates at https://validator.schema.org on a live site

### DIFF
--- a/Sources/Ignite/Elements/StructuredData.swift
+++ b/Sources/Ignite/Elements/StructuredData.swift
@@ -1,0 +1,392 @@
+//
+// StructuredData.swift
+// Ignite
+// https://www.github.com/twostraws/Ignite
+// See LICENSE for license information.
+//
+
+import Foundation
+
+/// A head element that renders JSON-LD structured data.
+///
+/// Use `StructuredData` to add Schema.org (or any JSON-LD) structured data
+/// to your pages. You can create schemas from a type and properties dictionary,
+/// from a raw JSON string, or use the built-in convenience methods for
+/// common schema types.
+///
+/// ```swift
+/// // Generic — any schema type
+/// Head {
+///     StructuredData("LocalBusiness", properties: [
+///         "name": "Joe's Pizza",
+///         "telephone": "555-0123"
+///     ])
+/// }
+///
+/// // Convenience methods
+/// Head {
+///     StructuredData.organization(name: "Acme", url: "https://acme.com")
+///     StructuredData.breadcrumbs()
+///     StructuredData.article()
+/// }
+///
+/// // Raw JSON-LD
+/// Head {
+///     StructuredData(json: customJSONString)
+/// }
+/// ```
+@MainActor
+public struct StructuredData: HeadElement {
+    /// The standard set of control attributes for HTML elements.
+    public var attributes = CoreAttributes()
+
+    /// Whether this HTML belongs to the framework.
+    public var isPrimitive: Bool { true }
+
+    /// How the JSON-LD content is specified.
+    private enum Content {
+        /// A pre-built JSON string, rendered as-is.
+        case raw(String)
+
+        /// A schema type and properties, serialized at render time.
+        case schema(context: String, type: String, properties: [String: Any])
+
+        /// Auto-generated Article schema from the current article context.
+        case article(publisher: String?, publisherURL: String?)
+
+        /// Auto-generated BreadcrumbList from the current page context.
+        case breadcrumbs(homeName: String)
+    }
+
+    private let content: Content
+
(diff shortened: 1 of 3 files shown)


Sources/Ignite/Elements/StructuredData.swift
```diff
@@ -0,0 +1,378 @@
...
+///     StructuredData.breadcrumbs()
+///     StructuredData.article()
+/// }
+///
+/// // Raw JSON-LD
+/// Head {
+///     StructuredData(json: customJSONString)
+/// }
+/// ```
+public struct StructuredData: HeadElement, Sendable {
```
MrSkwiggs, line 38: Since this is specifically about JSON-LD which stands for Linked Data, maybe this makes more sense here?

```suggestion
public struct LinkedData: HeadElement, Sendable {
```

or even

```suggestion
public struct SchemaData: HeadElement, Sendable {
```

Sources/Ignite/Elements/StructuredData.swift
```diff
@@ -0,0 +1,378 @@
...
+    ///
+    /// - Parameters:
+    ///   - name: The organization name.
+    ///   - url: The organization's website URL.
+    ///   - description: A brief description of the organization.
+    ///   - foundingDate: The founding date in ISO 8601 format (e.g., "2000-06-06").
+    ///   - sameAs: URLs for the organization's social media profiles or external pages.
+    ///   - parentOrganization: An optional parent organization as (name, url).
+    ///   - parentOrganizationType: The Schema.org type for the parent. Defaults to "Organization".
+    public static func organization(
```
MrSkwiggs, line 194: Since we're providing a few convenience constructors here, would it make sense to provide full parity from the get-go? IMHO if we introduce this feature with limited support, it wouldn't be quite user-friendly.

Now with that said, peeking at the amount of different types there are, maybe we could support most top-level constructors at least?

MrSkwiggs (commented):
I like the work done here, but I'm a bit torn with adding this and supporting this directly from within Ignite. 

I'm starting to think it might actually make more sense to vend this as a separate plugin. That way we could make the plugin exhaustive and offer full parity with Schema.org without overloading Ignite directly?

Something akin to how Vapor registers additional functionality might be interesting to look into, but yeah... that'd require quite a significant rework to properly support.

